PR #2, opened 2022-01-04
Set max method params count to 7

solid-yuriiprykhodko:


lib/analysis_options.yaml
```diff
@@ -20,8 +20,9 @@ dart_code_metrics:
   metrics:
     # NIST 500-235 item 2.5
     cyclomatic-complexity: 10
-    number-of-arguments: 4
```
solid-yuriiprykhodko, line 23: This option is the same as `number-of-parameters` in an old version of Dart Code Metrics: https://github.com/dart-code-checker/dart-code-metrics/blob/9b79547fffedeabf0d69f0725ec7db36b72dd99a/CHANGELOG.md#300-nullsafety0

solid-software (approved):

PR #7, opened 2022-01-11
add more useful rules

illia-romanenko:
migrate to dcm 4.9

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,9 @@ Then you can see suggestions in your IDE or you can run checks manually:
 
 ```bash
 dart analyze;
-dart run dart_code_metrics:metrics analyze lib;
-dart run dart_code_metrics:metrics analyze test;
-dart run dart_code_metrics:metrics check-unused-files lib;
-dart run dart_code_metrics:metrics check-unused-l10n lib;
+dart run dart_code_metrics:metrics analyze lib test;
+dart run dart_code_metrics:metrics check-unused-files lib test;
+dart run dart_code_metrics:metrics check-unused-l10n lib test;
 
 ```
 Beware that some of the `dart_code_metrics` checks are not displayed in IDE so running checks 
--- a/README.md
+++ b/README.md
@@ -30,10 +30,13 @@ Then you can see suggestions in your IDE or you can run checks manually:
 ```bash
 dart analyze;
 dart run dart_code_metrics:metrics analyze lib;
+dart run dart_code_metrics:metrics analyze test;
 dart run dart_code_metrics:metrics check-unused-files lib;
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -30,10 +30,13 @@ Then you can see suggestions in your IDE or you can run checks manually:
 ```bash
 dart analyze;
 dart run dart_code_metrics:metrics analyze lib;
+dart run dart_code_metrics:metrics analyze test;
 dart run dart_code_metrics:metrics check-unused-files lib;
 dart run dart_code_metrics:metrics check-unused-l10n lib;
```
solid-yuriiprykhodko, line 35:
```suggestion
dart run dart_code_metrics:metrics analyze lib test;
dart run dart_code_metrics:metrics check-unused-files lib test;
dart run dart_code_metrics:metrics check-unused-l10n lib test;
```

lib/analysis_options.yaml
```diff
@@ -30,12 +30,31 @@ dart_code_metrics:
...
+          - methods
     - newline-before-return
+    - no-empty-block
+    - no-equal-then-else
     - no-magic-number
+    - prefer-conditional-expressions
+    - prefer-first
+    - prefer-last
+    # This improves navigation by matching file content and file name.
+    - prefer-match-file-name
```
solid-yuriiprykhodko, line 57: 🥳

lib/analysis_options_test.yaml
```diff
@@ -48,6 +48,8 @@ dart_code_metrics:
 #      many places inside the test code, adding uninformative visual noise.
 #    - The use of this operator is also discouraged by the main ruleset.
     avoid-late-keyword: false
+# In test files it's ok to place stubs or other helper classes.
```
solid-yuriiprykhodko, line 51: Just some different wording, which to me seems a tiny bit more clear.
```suggestion
# It acceptable to include stubs or other helper classes into the main test file.
```

README.md
solid-yuriiprykhodko: Maybe we also should mention that `example` folder should be checked explicitly. 
Not sure we want to include it here -- it'll probably break the command for any project that doesn't have an example. But probably worth mentioning as a note below these commands.

illia-romanenko: Great suggestion! Thanks!

illia-romanenko: I think covering `example` maybe too much and will add more complexity to the docs? As it's a separate app it's generally makes sense that it's not recursive. Another case could be monorepo - we'll probably adjust for that case soon too? ;)

solid-yuriiprykhodko (approved):
Awesome additions -- thanks @illia-romanenko!

PR #14, opened 2022-09-16
add cast_nullable_to_non_nullable lint

solid-danylosafonov:
**Rationale:** this lint is similar to `avoid-non-null-assertion` which is already used, basically it suggests to not ignore possible `null` value


solid-software (changes requested):
Great job! Could you please add Rationale to yaml itself (in the comments above the rule) so that it can be used to understand why this rule is enabled?

solid-software (approved):
LGTM - thanks!

PR #15, opened 2022-10-10
use_colored_box & use_decorated_box

solid-danylosafonov:


lib/analysis_options.yaml
```diff
@@ -244,8 +244,8 @@ linter:
     - unrelated_type_equality_checks
     - unsafe_html
     - use_build_context_synchronously
-    # available in later releases
```
illia-romanenko, line 247: Could you please check whether this rule is available in the linked release of analyzer?

solid-danylosafonov: As I understand version of analyser (2.8.0 in our case) doesn't matter (correct me if I'm wrong).
Because lints are defined in linter package, which should be ^1.19, but it is bundled with sdk so what really matters is sdk version - ^2.14.4 in our case, which is enough to use provided lints because linter uses sdk ^2.12 up to version 1.21.0. I will also update some of the comments "available in later releases" to include the version of linter (all of them can be used with current sdk - should I uncomment them?)

PR #19, opened 2023-01-26
Disable deprecated rule

solid-oleksandrhordiienko:


lib/analysis_options.yaml
```diff
@@ -146,7 +146,8 @@ linter:
     - file_names
     - hash_and_equals
     - implementation_imports
-    - invariant_booleans
+    # deprecated rule
```
illia-romanenko, line 149:
```suggestion
    # deprecated rule in Flutter 3.7
```

illia-romanenko (changes requested):
Thanks for the PR, please also add changelog and increment version. Thanks!

illia-romanenko (approved):
LGTM - thanks!

PR #26, opened 2023-08-09
Update Dart SDK version constraints

n-bernat:
Bump SDK version to Dart 3 in order to allow for smooth transition from `dart_code_metrics` to `custom_lint` (which is required for `custom_lint` package).


solid-yuriiprykhodko (approved):
LGTM.
Let's work in a feature branch for now.

PR #29, opened 2023-08-09
Release v0.1.0 preview

### LINKED ISSUE #28: Migrate from DCM to custom_lint package
As the [open-source DCM is discontinued](https://pub.dev/packages/dart_code_metrics) we have to migrate to a different solution. `custom_lint` package seems like a good option as it allows us to create custom lints quite easily.

This issue is here to organise the progress of migrating from `dart_code_metrics` to `custom_lint`. There are a number of rules that have to be recreated before full migration. Some of them may have appeared as a standard lint, so they just have to be moved to a section that doesn't depend on DCM. Checklist below is divided based on the DCM config file.

Metrics:
- [x] cyclomatic-complexity
- [x] number-of-parameters
- [x] source-lines-of-code

Rules:
- [x] avoid-global-state
- [x] avoid-late-keyword
- [x] avoid-non-null-assertion
- [x] avoid-returning-widgets
- [x] avoid-unnecessary-setstate
- [x] avoid-unnecessary-type-assertions
- [x] avoid-unnecessary-type-casts
- [x] avoid-unrelated-type-assertions
- [x] avoid-unused-parameters
- [x] double-literal-format
- [x] member-ordering (custom format)
- [x] newline-before-return
- [x] no-empty-block
- [x] no-equal-then-else
- [x] no-magic-number
- [x] prefer-conditional-expressions
- [x] prefer-first
- [x] prefer-last
- [x] prefer-match-file-name

Anti-patterns:
- [ ] long-method
- [ ] long-parameter-list

solid-yuriiprykhodko:
Closes #28


illia-romanenko:
Great job! Thanks everybody!

PR #31, opened 2023-08-09
Add number of parameters metric

n-bernat:
PR which allows the capability to check whether we excedeed the maximum allowed number of parameters. As a first new lint it also contains a proposal for organising the code. 

**One thing to consider**: In my opinion having custom models for lints is not necessary, operating on primitives like:
```yaml
custom_lint:
  rules:
    - cyclomatic_complexity_metric: 4
    - number_of_parameters: 2
```
should be good enough for us.

Current configuration for custom lints looks like that:
```yaml
custom_lint:
  rules:
    - cyclomatic_complexity_metric:
      max_complexity: 4
    - number_of_parameters:
      max_parameters: 2
```
and requires custom models for each rule. It feels like an overkill for me, because we will end up with a bunch of models that have only one field in them.

### DIFF
--- a/lib/number_of_parameters/models/number_of_parameters_parameters.dart
+++ b/lib/number_of_parameters/models/number_of_parameters_parameters.dart
@@ -1,0 +1,19 @@
+/// A data model class that represents the "number of parameters" input
+/// parameters.
+class NumberOfParametersParameters {
+  /// Maximum number of parameters
+  final int maxParameters;
+
+  static const _defaultMaxParameters = 2;
+
+  /// Constructor for [NumberOfParametersParameters] model
+  const NumberOfParametersParameters({
+    required this.maxParameters,
+  });
+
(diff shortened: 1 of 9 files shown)


lib/number_of_parameters/models/number_of_parameters_parameters.dart
```diff
@@ -0,0 +1,19 @@
+/// A data model class that represents the "number of parameters" input
+/// parameters.
+class NumberOfParametersParameters {
+  /// Maximum number of parameters
+  final int maxParameters;
+
+  static const _defaultMaxParameters = 2;
```
solid-yuriiprykhodko, line 7: Might make sense co collect all defaults in one place later on. For now let's roll with local constantw

solid-yuriiprykhodko (approved):
Nice and concise.

PR #32, opened 2023-08-10
0.0.19

solid-danylosafonov:


solid-danylosafonov:
@illia-romanenko

lib/analysis_options.yaml
```diff
@@ -252,18 +255,18 @@ linter:
...
-    # linter v1.19+
     - use_colored_box
-    # linter v1.15+
     - use_decorated_box
     - use_full_hex_values_for_flutter_colors
     - use_function_type_syntax_for_parameters
     - use_if_null_to_convert_nulls_to_bools
     - use_is_even_rather_than_modulo
     - use_named_constants
+    - use_raw_strings
```
illia-romanenko, line 269: This could potentially lead to annoyances like when you would want to add interpolation and it will not work. It feels like maybe it doesn't need to be enforced in the end.

illia-romanenko (changes requested):
Great job! Thanks for putting it together. Added some thoughts and feel free to ping or discuss it further.

lib/analysis_options.yaml
```diff
@@ -197,6 +199,7 @@ linter:
     - prefer_final_fields
     - prefer_final_in_for_each
     - prefer_final_locals
+    - prefer_foreach
```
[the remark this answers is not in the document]
solid-danylosafonov: I think it is similar to [unnecessary_lambdas](https://dart.dev/tools/linter-rules/unnecessary_lambdas)


BAD:
```dart
// unnecessary_lambdas
f(callback: (x) => callback(x));

// prefer_foreach
for (final x in xs) {
  callback(x);
} 
```

GOOD:
```dart
// unnecessary_lambdas
f(callback: callback);

// prefer_foreach
xs.forEach(callback);
```

lib/analysis_options.yaml
solid-danylosafonov: I'm not sure if it can lead to issues like that
e.g. this doesn't trigger the lint:
```dart
RegExp('${key}_(\\d+)u_1(m|y)\$')
```

I think it is safe to assume that it only triggers when escaped symbols aren't used unescaped in the same string

This will trigger:
```dart
RegExp('key_(\\d+)u_1(m|y)\$')
```

illia-romanenko: what I mean if we force raw string and then would want to add interpolation (after some time).

solid-yuriiprykhodko: If we have a raw string somewhere:

```dart
final s = r'some complex text with different " 'quotes" ' and $ signs';
```

And then want to interpolate something there, we'll be forced to add escaping everywhere, instead of simply doing what we want:

```dart
final s = 'some complex ${someVar} text with different \" \'quotes\" \' and \$ signs';
```

This will make for non-clean diffs, and could be error-prone.

solid-yuriiprykhodko: Besides, even my example above shows some issues with quote handling in raw strings.

solid-danylosafonov: This example doesn't trigger the lint, but I understand the issue with the diffs
```dart
'some complex text with different " \'quotes" \' and \$ signs';
```

solid-danylosafonov: > This example doesn't trigger the lint

As in it won't force us to use a raw-string instead of a normal one here

solid-danylosafonov: even this doesn't trigger a lint
```dart
'some complex text with different \'quotes \' and \$ signs';
```

(it can actually be a raw string):
```dart
r"some complex text with different 'quotes ' and $ signs";
```

solid-danylosafonov: As for the diff issue - isn't the code read much more often than it is written?
IMO raw strings improve readability
Even if it is edited a few times after it was written, it will be probably read more times

WDYT?

lib/analysis_options.yaml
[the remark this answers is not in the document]
solid-danylosafonov: Okay, I'll comment it out and add an explanation then

illia-romanenko (approved):
LGTM - thanks!

PR #34, opened 2023-08-10
Add linter for lines of code

n-bernat:


lib/lints/lines_of_code/lines_of_code_metric.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:solid_lints/lints/lines_of_code/models/lines_of_code_parameters.dart';
+import 'package:solid_lints/models/metric_rule.dart';
+import 'package:solid_lints/visitors/source_code_visitor.dart';
+
+/// A number of lines metric which checks whether we didn't exceed
+/// the maximum allowed number of lines for a file
+class LinesOfCodeMetric extends DartLintRule {
```
solid-yuriiprykhodko, line 9: For a file or for a function?
Originally Dart Code Metrics counts lines per function. So we absolutely can have a class with >200 LoC.

solid-yuriiprykhodko: https://github.com/dart-code-checker/dart-code-metrics/blob/master/lib/src/analyzers/lint_analyzer/metrics/metrics_list/lines_of_code_metric.dart#L24

lib/visitors/source_code_visitor.dart
```diff
@@ -0,0 +1,69 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/source/line_info.dart';
+
+/// The AST visitor that will find lines with code.
+class SourceCodeVisitor extends RecursiveAstVisitor<void> {
```
solid-yuriiprykhodko, line 30:
```suggestion
class LinesOfCodeVisitor extends RecursiveAstVisitor<void> {
```

lib/lints/lines_of_code/lines_of_code_metric.dart
solid-yuriiprykhodko: Yeah, agree that the naming is misleading. Something like FunctionLinesOfCode and function_lines_of_code would be better.

solid-yuriiprykhodko (approved):
LGTM

PR #35, opened 2023-08-10
Add missing tests for custom_lint

n-bernat:
`// expect_lint` allows us to specify that on the following line there should be a match. It returns an error if the line below that is not matched. If it's matched then this line is ignored.


solid-yuriiprykhodko (approved):
LGTM, good find

PR #36, opened 2023-08-10
Add avoid_global_state lint

n-bernat:
Those `MetricRule` definitions should probably be moved to a separate place later on (maybe to associated lints?). It takes too much space in the main file considering how many lints we have to implement. 

cc: @solid-yuriiprykhodko

### DIFF
--- a/lib/lints/avoid_global_state/avoid_global_state_rule.dart
+++ b/lib/lints/avoid_global_state/avoid_global_state_rule.dart
@@ -1,0 +1,28 @@
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+
+/// A global state rule which forbids using variables
+/// that can be globally modified.
+class AvoidGlobalStateRule extends DartLintRule {
+  /// The [LintCode] of this lint rule that represents
+  /// the error whether we use global state.
+  static const lintName = 'avoid_global_state';
+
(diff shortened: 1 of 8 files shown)


lib/lints/avoid_global_state/avoid_global_state_rule.dart
```diff
@@ -0,0 +1,28 @@
...
+    CustomLintResolver resolver,
+    ErrorReporter reporter,
+    CustomLintContext context,
+  ) {
+    context.registry.addVariableDeclaration((node) {
+      final isPrivate = node.declaredElement?.isPrivate ?? false;
+
+      if (!isPrivate && !node.isFinal && !node.isConst) {
+        reporter.reportErrorForNode(code, node);
+      }
```
solid-yuriiprykhodko, line 25: So we're allowing private global vars? Is this how it was implemented in DCM?

solid-yuriiprykhodko: Hm, I wonder what the rationale was for it.

For one, we wouldn't be able to implement Singletons if both private vars and `late` are forbidden. And good riddance, to be honestly :D.

Let's keep it this way. We can modify later if we collect more data on the _why_.

solid-yuriiprykhodko (approved):
LGTM. Please resolve conflicts so I can merge.

PR #40, opened 2023-08-11
Add avoid_returning_widgets lint

nikodembernat:
cc: @solid-yuriiprykhodko

### DIFF
--- a/lib/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
+++ b/lib/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
@@ -20,8 +20,7 @@ class AvoidReturningWidgetsRule extends DartLintRule {
   ) {
     context.registry.addDeclaration((node) {
       final isWidgetReturned = switch (node) {
-        FunctionDeclaration(returnType: TypeAnnotation(:final type?)) =>
-          hasWidgetType(type),
+        FunctionDeclaration(returnType: TypeAnnotation(:final type?)) ||
         MethodDeclaration(returnType: TypeAnnotation(:final type?)) =>
           hasWidgetType(type),
         _ => false,


example/test/avoid_returning_widget_test.dart
```diff
@@ -0,0 +1,31 @@
...
+Widget avoidReturningWidgets() => const SizedBox();
+
+class MyWidget extends StatelessWidget {
+  const MyWidget({super.key});
+
+  // expect_lint: avoid_returning_widgets
+  Widget _test1() => const SizedBox();
+
+  // expect_lint: avoid_returning_widgets
+  Widget _test2() {
```
solid-yuriiprykhodko, line 18: Does this produce a lint too?
```suggestion
  dynamic _test2() {
```

lib/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
```diff
@@ -0,0 +1,46 @@
...
+    context.registry.addDeclaration((node) {
+      final bool isWidgetReturned;
+      if (node is FunctionDeclaration) {
+        final returnType = node.returnType?.type;
+        isWidgetReturned = returnType != null && hasWidgetType(returnType);
+      } else if (node is MethodDeclaration) {
+        // `build` methods return widgets by nature
+        if (node.declaredElement?.name == "build") {
+          return;
+        }
```
solid-yuriiprykhodko, line 30: Can we use Dart pattern matching to simplify this?

lib/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
```diff
@@ -20,22 +20,23 @@ class AvoidReturningWidgetsRule extends DartLintRule {
...
-        isWidgetReturned = returnType != null && hasWidgetType(returnType);
-      } else {
-        isWidgetReturned = false;
+          // `FunctionDeclaration` and `MethodDeclaration` implement
+          // different things to have this `returnType` getter so
+          // unfortunately it can't be simplified.
+          final returnType = node.returnType?.type;
+          isWidgetReturned = returnType != null && hasWidgetType(returnType);
+        default:
+          isWidgetReturned = false;
```
solid-yuriiprykhodko, line 39: I was thinking something like this. Cuts down a few null checks but looks... not quite readable.

@n-bernat what do you think.
```suggestion
      switch (node) {
        case (FunctionDeclaration(
            returnType: TypeAnnotation(:final type?),
          )):
          isWidgetReturned = hasWidgetType(type);
        case (MethodDeclaration(
            returnType: TypeAnnotation(:final type?),
            declaredElement: ExecutableElement(:final name)
          )):
          // `build` methods return widgets by nature
          if (name == "build") return;

          // `FunctionDeclaration` and `MethodDeclaration` implement
          // different things to have this `returnType` getter so
          // unfortunately it can't be simplified.
          isWidgetReturned = hasWidgetType(type);
        default:
          isWidgetReturned = false;
```

example/test/avoid_returning_widget_test.dart
solid-yuriiprykhodko: Not in scope now. Right now the priority is to get off DCM, we should focus that.
This probably worked the same in DCM, also.

lib/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
solid-yuriiprykhodko: That's actually quite good. A lot less clutter, nice and to the point. Let's do this.

lib/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
```diff
@@ -0,0 +1,36 @@
...
+    CustomLintResolver resolver,
+    ErrorReporter reporter,
+    CustomLintContext context,
+  ) {
+    context.registry.addDeclaration((node) {
+      final isWidgetReturned = switch (node) {
+        FunctionDeclaration(returnType: TypeAnnotation(:final type?)) =>
+          hasWidgetType(type),
+        MethodDeclaration(returnType: TypeAnnotation(:final type?)) =>
+          hasWidgetType(type),
```
solid-yuriiprykhodko, line 26: ?
```suggestion
        FunctionDeclaration(returnType: TypeAnnotation(:final type?)) ||
        MethodDeclaration(returnType: TypeAnnotation(:final type?)) =>
          hasWidgetType(type),
```

solid-yuriiprykhodko: I'm still a bit reserved against pattern matching, but honestly -- this looks pretty good in comparison with the initial classic approach.

solid-yuriiprykhodko (approved):
LGTM!

PR #41, opened 2023-08-11
Cleanup of lint creation

n-bernat:
As we have more and more lints it's a good idea to split it, so we won't end up with a one giant method with all lints.

cc: ~illia-romanenko~ it was supposed to be Yurii

### DIFF
--- a/lib/models/solid_lint_rule.dart
+++ b/lib/models/solid_lint_rule.dart
@@ -1,0 +1,16 @@
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:solid_lints/models/metric_rule.dart';
+
+/// A base class for emitting information about
+/// issues with user's `.dart` files.
+abstract class SolidLintRule<T extends Object?> extends DartLintRule {
+  /// Constructor for [SolidLintRule] model.
+  SolidLintRule(this.config) : super(code: config.lintCode);
+
+  /// Configuration for a particular rule with all the
+  /// defined custom parameters.
+  final MetricRule<T> config;
+
+  /// A flag which indicates whether this rule was enabled by the user.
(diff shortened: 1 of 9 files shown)


lib/models/solid_lint_rule.dart
```diff
@@ -0,0 +1,16 @@
...
+
+/// A base class for emitting information about
+/// issues with user's `.dart` files.
+abstract class SolidLintRule<T extends Object?> extends DartLintRule {
+  /// Constructor for [SolidLintRule] model.
+  SolidLintRule(this.config) : super(code: config.lintCode);
+
+  /// Configuration for a particular rule with all the
+  /// defined custom parameters.
+  final MetricRule<T> config;
```
solid-yuriiprykhodko, line 12: Maybe we should merge SolidLintRule and MetricRule into a single entity? This would simplify the declaration of these rules also -- it will just be an override on the rule implementation.

lib/solid_lints.dart
```diff
@@ -6,111 +6,28 @@ import 'package:solid_lints/lints/avoid_late_keyword/avoid_late_keyword_rule.dar
...
+      NumberOfParametersMetric.createRule(configs),
+      FunctionLinesOfCodeMetric.createRule(configs),
+      AvoidNonNullAssertionRule.createRule(configs),
+      AvoidLateKeywordRule.createRule(configs),
+      AvoidGlobalStateRule.createRule(configs),
+      AvoidReturningWidgetsRule.createRule(configs),
+    ];
+
+    // Return only enabled rules
+    return supportedRules.where((r) => r.enabled).toList();
```
solid-yuriiprykhodko, line 31: Neat!

PR #46, opened 2023-09-01
Feature/avoid unnecessary type assertions

xvadim:
1. Implemented the avoid-unnecessary-type-assertions rule
2. Implemented quick-fixes for the avoid-unnecessary-type-assertions rule

### DIFF
--- a/example/test/avoid_unnecessary_type_assertions_test.dart
+++ b/example/test/avoid_unnecessary_type_assertions_test.dart
@@ -13,11 +13,20 @@ void fun() {
   // expect_lint: avoid-unnecessary-type-assertions
   testList.whereType<double>();
 
+  final dynamicList = <dynamic>[1.0, 2.0];
+  dynamicList.whereType<double>();
+
+  // just any method, except `whereType`
+  testList.add(4.0);
+
+  // expect_lint: avoid-unnecessary-type-assertions
+  [1.0, 2.0].whereType<double?>();
+
   final double d = 2.0;
   // expect_lint: avoid-unnecessary-type-assertions
   final casted = d is double;
 
   final double? nullableD = 2.0;
-  // `Type? is Type` is allowed
+  // casting `Type? is Type` is allowed
   final castedD = nullableD is double;
 }
--- a/example/test/avoid_unnecessary_type_assertions_test.dart
+++ b/example/test/avoid_unnecessary_type_assertions_test.dart
@@ -10,8 +10,9 @@ void fun() {
   // expect_lint: avoid-unnecessary-type-assertions
   final result = testList is List<double>;
 
+  // to check quick-fix => testList.length
   // expect_lint: avoid-unnecessary-type-assertions
-  testList.whereType<double>();
+  testList.whereType<double>().length;
 
   final dynamicList = <dynamic>[1.0, 2.0];
   dynamicList.whereType<double>();
--- a/example/test/avoid_unnecessary_type_assertions_test.dart
+++ b/example/test/avoid_unnecessary_type_assertions_test.dart
@@ -1,6 +1,7 @@
-// ignore_for_file: unnecessary_nullable_for_final_variable_declarations
 // ignore_for_file: prefer_const_declarations
+// ignore_for_file: unnecessary_nullable_for_final_variable_declarations
 // ignore_for_file: unnecessary_type_check
+// ignore_for_file: unused_local_variable
 
 /// Check the `avoid-unnecessary-type-assertions` rule
 
--- a/example/test/avoid_unnecessary_type_assertions_test.dart
+++ b/example/test/avoid_unnecessary_type_assertions_test.dart
@@ -1,0 +1,22 @@
+// ignore_for_file: unnecessary_nullable_for_final_variable_declarations
+// ignore_for_file: prefer_const_declarations
+// ignore_for_file: unnecessary_type_check
+
+/// Check the `avoid-unnecessary-type-assertions` rule
+
+void fun() {
+  final testList = <double>[1.0, 2.0, 3.0];
+  // expect_lint: avoid-unnecessary-type-assertions
+  final result = testList is List<double>;
+
+  // expect_lint: avoid-unnecessary-type-assertions
+  testList.whereType<double>();
+
+  final double d = 2.0;
+  // expect_lint: avoid-unnecessary-type-assertions
+  final casted = d is double;
+
+  final double? nullableD = 2.0;
+  // `Type? is Type` is allowed
+  final castedD = nullableD is double;
+}
--- a/example/test/avoid_unnecessary_type_assertions_test.dart
+++ b/example/test/avoid_unnecessary_type_assertions_test.dart
@@ -6,9 +6,12 @@
 /// Check the `avoid-unnecessary-type-assertions` rule
 
 void fun() {
-  final testList = <double>[1.0, 2.0, 3.0];
+  final testList = [1.0, 2.0, 3.0];
   // expect_lint: avoid-unnecessary-type-assertions
   final result = testList is List<double>;
+
+  // expect_lint: avoid-unnecessary-type-assertions
+  final negativeResult = testList is! List<double>;
 
   // to check quick-fix => testList.length
   // expect_lint: avoid-unnecessary-type-assertions
@@ -17,15 +20,15 @@ void fun() {
   final dynamicList = <dynamic>[1.0, 2.0];
   dynamicList.whereType<double>();
 
-  // just any method, except `whereType`
-  testList.add(4.0);
-
   // expect_lint: avoid-unnecessary-type-assertions
   [1.0, 2.0].whereType<double?>();
 
   final double d = 2.0;
   // expect_lint: avoid-unnecessary-type-assertions
   final casted = d is double;
+
+  // expect_lint: avoid-unnecessary-type-assertions
+  final negativeCasted = d is! double;
 
   final double? nullableD = 2.0;
   // casting `Type? is Type` is allowed
--- a/lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
+++ b/lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
@@ -13,7 +13,7 @@ class _UnnecessaryTypeAssertionsFix extends DartFix {
   ) {
     context.registry.addIsExpression((node) {
       if (analysisError.sourceRange.intersects(node.sourceRange)) {
-        _addDeletion(reporter, 'is', node, node.isOperator.offset);
+        _addDeletion(reporter, operatorIsName, node, node.isOperator.offset);
       }
     });
 
@@ -21,7 +21,7 @@ class _UnnecessaryTypeAssertionsFix extends DartFix {
       if (analysisError.sourceRange.intersects(node.sourceRange)) {
         _addDeletion(
           reporter,
-          'whereType',
+          whereTypeMethodName,
           node,
           node.operator?.offset ?? node.offset,
         );
--- a/lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
+++ b/lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
@@ -12,43 +12,39 @@ class _UnnecessaryTypeAssertionsFix extends DartFix {
     List<AnalysisError> others,
   ) {
     context.registry.addIsExpression((node) {
-      // checks that the literal declaration is where our warning is located
-      if (!analysisError.sourceRange.intersects(node.sourceRange)) return;
-
-      final changeBuilder = reporter.createChangeBuilder(
-        message: "Remove unnecessary 'is'",
-        priority: 1,
-      );
-
-      final isOperatorOffset = node.isOperator.offset;
-      changeBuilder.addDartFileEdit((builder) {
-        builder.addDeletion(
-          SourceRange(
-            isOperatorOffset,
-            node.length - (isOperatorOffset - node.offset),
-          ),
-        );
-      });
+      if (analysisError.sourceRange.intersects(node.sourceRange)) {
+        _addDeletion(reporter, 'is', node, node.isOperator.offset);
+      }
     });
 
     context.registry.addMethodInvocation((node) {
-      // checks that the literal declaration is where our warning is located
-      if (!analysisError.sourceRange.intersects(node.sourceRange)) return;
-
-      final changeBuilder = reporter.createChangeBuilder(
-        message: "Remove unnecessary 'whereType'",
-        priority: 1,
-      );
-
-      final operatorOffset = node.operator?.offset ?? node.offset;
-      changeBuilder.addDartFileEdit((builder) {
-        builder.addDeletion(
-          SourceRange(
-            operatorOffset,
-            node.length - (operatorOffset - node.offset),
-          ),
+      if (analysisError.sourceRange.intersects(node.sourceRange)) {
+        _addDeletion(
+          reporter,
+          'whereType',
+          node,
+          node.operator?.offset ?? node.offset,
         );
-      });
+      }
+    });
+  }
+
+  void _addDeletion(
+    ChangeReporter reporter,
+    String itemToDelete,
+    Expression node,
+    int operatorOffset,
+  ) {
+    final targetNameLength = operatorOffset - node.offset;
+    final removedPartLength = node.length - targetNameLength;
+
+    final changeBuilder = reporter.createChangeBuilder(
+      message: "Remove unnecessary '$itemToDelete'",
+      priority: 1,
+    );
+
+    changeBuilder.addDartFileEdit((builder) {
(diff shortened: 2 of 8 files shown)


example/test/avoid_unnecessary_type_assertions_test.dart
```diff
@@ -0,0 +1,33 @@
...
+// ignore_for_file: unnecessary_nullable_for_final_variable_declarations
+// ignore_for_file: unnecessary_type_check
+// ignore_for_file: unused_local_variable
+
+/// Check the `avoid-unnecessary-type-assertions` rule
+
+void fun() {
+  final testList = <double>[1.0, 2.0, 3.0];
+  // expect_lint: avoid-unnecessary-type-assertions
+  final result = testList is List<double>;
```
solid-yuriiprykhodko, line 11: Not sure we need to be explicit here. Let type inference do its job.

```suggestion
  final testList = [1.0, 2.0, 3.0];
  // expect_lint: avoid-unnecessary-type-assertions
  final result = testList is List<double>;
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
```diff
@@ -0,0 +1,54 @@
...
+        message: "Remove unnecessary 'is'",
+        priority: 1,
+      );
+
+      final isOperatorOffset = node.isOperator.offset;
+      changeBuilder.addDartFileEdit((builder) {
+        builder.addDeletion(
+          SourceRange(
+            isOperatorOffset,
+            node.length - (isOperatorOffset - node.offset),
```
solid-yuriiprykhodko, line 28: Not sure we need the parentheses.
A better option might be to put the result of this calculation into a properly named variable.
```suggestion
            node.length - isOperatorOffset - node.offset,
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
```diff
@@ -0,0 +1,54 @@
...
+
+      final operatorOffset = node.operator?.offset ?? node.offset;
+      changeBuilder.addDartFileEdit((builder) {
+        builder.addDeletion(
+          SourceRange(
+            operatorOffset,
+            node.length - (operatorOffset - node.offset),
+          ),
+        );
+      });
```
solid-yuriiprykhodko, line 51: Seems exactly the same as above.
Extract method?

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+  static const lintName = 'avoid-unnecessary-type-assertions';
+
+  static const _unnecessaryIsCode = LintCode(
+    name: lintName,
+    problemMessage: "Unnecessary usage of the 'is' operator.",
+  );
+
+  static const _unnecessaryWhereTypeCode = LintCode(
+    name: lintName,
+    problemMessage: "Unnecessary usage of the 'whereType'.",
```
solid-yuriiprykhodko, line 29:
```suggestion
    problemMessage: "Unnecessary usage of the 'whereType' method.",
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+  AvoidUnnecessaryTypeAssertions._(super.config);
+
+  /// Creates a new instance of [AvoidUnnecessaryTypeAssertions]
+  /// based on the lint configuration.
+  factory AvoidUnnecessaryTypeAssertions.createRule(CustomLintConfigs configs) {
+    final rule = RuleConfig(
+      configs: configs,
+      name: lintName,
+      problemMessage: (_) =>
+          "Unnecessary usage of 'is' or 'whereType' operators.",
```
solid-yuriiprykhodko, line 41: Generalize. Especially given we're going to expand this rule with `case` statements support later on.
```suggestion
          "Unnecessary usage of typecast operators.",
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  @override
+  List<Fix> getFixes() => [_UnnecessaryTypeAssertionsFix()];
+
+  bool _isUnnecessaryIsExpression(IsExpression node) {
+    final objectType = node.expression.staticType;
+    final castedType = node.type.type;
+
+    if (node.notOperator != null) {
+      if (objectType != null &&
```
solid-yuriiprykhodko, line 74:
```suggestion
    if (node.notOperator != null &&
        objectType != null &&
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  bool _isUnnecessaryWhereType(MethodInvocation node) {
+    const whereTypeMethodName = 'whereType';
+
+    final targetType = node.target?.staticType;
+
+    if (node.methodName.name == whereTypeMethodName &&
+        isIterableOrSubclass(node.realTarget?.staticType) &&
+        targetType is ParameterizedType) {
+      final isTargetTypeHasGeneric = targetType.typeArguments.isNotEmpty;
```
solid-yuriiprykhodko, line 97: The fact that it's generic is established above. This checks for _actual_ presence of explicit type args.
```suggestion
      final targetHasExplicitTypeArgs = targetType.typeArguments.isNotEmpty;
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+      final arguments = node.typeArguments?.arguments;
+      final isWhereTypeHasGeneric = arguments?.isNotEmpty ?? false;
```
solid-yuriiprykhodko, line 99:
```suggestion
      final invocationHasExplicitTypeArgs = arguments?.isNotEmpty ?? false;
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+          castedType,
+          isReversed: true,
+        );
+      }
+    }
+
+    return _isUnnecessaryTypeCheck(objectType, castedType);
+  }
+
+  bool _isUnnecessaryWhereType(MethodInvocation node) {
```
solid-yuriiprykhodko, line 89: Maybe we can simplify this whole method by using pattern matching?

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+    final isObjectTypeNullable = isNullableType(objectType);
+    final isCastedTypeNullable = isNullableType(castedType);
+
+    // Only one case `Type? is Type` always valid assertion case.
+    return isObjectTypeNullable && !isCastedTypeNullable;
+  }
+
+  DartType? _castedTypeInObjectTypeHierarchy(
+    DartType objectType,
+    DartType castedType,
```
solid-yuriiprykhodko, line 148:
```suggestion
  DartType? _castTypeInHierarchy(
    DartType objectType,
    DartType castType,
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+
+    if (objectType is InterfaceType) {
+      return objectType.allSupertypes
+          .firstWhereOrNull((value) => value.element == castedType.element);
+    }
+
+    return null;
+  }
+
+  bool _isGenerics(DartType objectType, DartType castedType) {
```
solid-yuriiprykhodko, line 162: The naming doesn't really help understand what the method is doing.

Maybe something like this?
```suggestion
  bool _areGenericsWithSameTypeArgs(DartType objectType, DartType castedType) {
```

solid-yuriiprykhodko, line 162: The naming doesn't really help understand what the method is doing.

Maybe something like this?
```suggestion
  bool _areGenericsWithSameTypeArgs(DartType objectType, DartType castedType) {
```

lib/utils/types_utils.dart
```diff
@@ -35,6 +36,12 @@ bool hasWidgetType(DartType type) =>
         _isListInheritedProvider(type) ||
         _isFutureInheritedProvider(type));
 
+bool isIterableOrSubclass(DartType? type) =>
+    _checkSelfOrSupertypes(type, (t) => t?.isDartCoreIterable ?? false);
```
solid-yuriiprykhodko, line 40: All subclasses of Iterable are also Iterable.
```suggestion
bool isIterable(DartType? type) =>
    _checkSelfOrSupertypes(type, (t) => t?.isDartCoreIterable ?? false);
```

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    for (var argumentIndex = 0; argumentIndex < length; argumentIndex++) {
+      if (!_isUnnecessaryTypeCheck(
+        objectType.typeArguments[argumentIndex],
+        castedType.typeArguments[argumentIndex],
+      )) {
+        return false;
+      }
+    }
```
solid-yuriiprykhodko, line 179: Maybe zip + every will look better here?

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
solid-yuriiprykhodko, line 148: Quite a few places where these go together. Maybe a small structure will make things simpler?

```dart
class TypeCast {
  final DartType source;
  final DartType target;
}
```

solid-yuriiprykhodko (changes requested):

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
xvadim: We need the parentheses here. But variables with proper names are better.

solid-yuriiprykhodko: Ah, right -- the grouping alters the sign of operation. So `a - (b - c)` -> `a + -(b + (-c))` -> `a + ((-b) - (-c))` -> `a + (- b + c)` -> `a - b + c`.

```suggestion
            node.length - isOperatorOffset + node.offset,
```

solid-yuriiprykhodko: BTW why is an integer variable named with a `isX` prefix?

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
xvadim: I am not sure that this variant will be simpler and more understable.

xvadim:
1. Fixed warnings
2. Fixed a bug
3. Added additional test cases

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
solid-yuriiprykhodko: A comparison would be helpful

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
solid-yuriiprykhodko: In the context of this lint, the typecast source and target are highly coupled.
In simple terms -- everywhere they go, they go together.

So if we're not using procedural/imperative style -- it looks a lot like a small struct. 

Some utility methods look like they would belong to this structure quite well -- anything that compares the two fields.

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
```diff
@@ -0,0 +1,50 @@
...
+      if (analysisError.sourceRange.intersects(node.sourceRange)) {
+        _addDeletion(reporter, 'is', node, node.isOperator.offset);
+      }
+    });
+
+    context.registry.addMethodInvocation((node) {
+      if (analysisError.sourceRange.intersects(node.sourceRange)) {
+        _addDeletion(
+          reporter,
+          'whereType',
```
solid-yuriiprykhodko, line 24: Don't you have it defined as a constant elsewhere?
Let's reuse a single definition instead of hardcoding strings.

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,165 @@
...
+            arguments?.first.type,
+          );
+    }
+    return false;
+  }
+
+  bool _isUnnecessaryTypeCheck(
+    DartType? objectType,
+    DartType? castedType, {
+    bool isReversed = false,
```
solid-yuriiprykhodko, line 109: I find it hard to understand what this flag does in the method.
Need better naming, code comments or functional decomposition.

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_fix.dart
xvadim: Because it is an offset of the 'is' operator. Maybe the better name is operatorIsOffset.

solid-yuriiprykhodko: `isOperatorIsOffset` -- sounds confusing
`operatorIsOffset` would make me double-take too.
Maybe `typeCheckOperatorOffset`?

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,188 @@
...
+    if (objectType is InterfaceType) {
+      return objectType.allSupertypes.firstWhereOrNull(
+        (value) => value.element == typeCast.target.element,
+      );
+    }
+
+    return null;
+  }
+
+  bool _areGenericsWithSameTypeArgs(TypeCast typeCast) {
```
solid-yuriiprykhodko, line 173: These unary methods look like they belong on TypeCast class.

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
```diff
@@ -0,0 +1,188 @@
...
+    if (typeCast.source is! ParameterizedType ||
+        typeCast.target is! ParameterizedType) {
+      return false;
+    }
+
+    final objectType = typeCast.source as ParameterizedType;
+    final castedType = typeCast.target as ParameterizedType;
```
solid-yuriiprykhodko, line 180: Will look better with pattern matching.

lib/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule.dart
xvadim: We can use something like this:
`
if (node
          case MethodInvocation(
            methodName: Identifier(name: 'whereType'),
            target: Expression(staticType: final targetType),
            realTarget: Expression(staticType: final realTargetType),
            typeArguments: TypeArgumentList(arguments: final arguments),
          )
          when targetType is ParameterizedType &&
              isIterable(realTargetType) &&
              arguments.isNotEmpty) {
        return _isUnnecessaryTypeCheck(
          targetType.typeArguments.first,
          arguments.first.type,
        );
      }

return false;
`

solid-yuriiprykhodko: Although I'm usually reserved against this stuff, it does look more readable than the classic approach. Nice work!

solid-yuriiprykhodko (approved):
LGTM

PR #47, opened 2023-09-13
Avoid unnecessary type casts rule

DenisBogatirov:
Added avoid-unnecessary-type-casts rule, with test and quick fix
Refactored avoid-unnecessary-type-assertion & types_utils, extracted common methods

### DIFF
--- a/lib/utils/typecast_utils.dart
+++ b/lib/utils/typecast_utils.dart
@@ -17,9 +17,16 @@ class TypeCast {
   /// The type being tested for
   final DartType target;
 
+  /// Set to true for opposite comparison, i.e 'is!'
+  final bool isReversed;
+
   /// Creates a new Typecast object with a given expression's or object's type
   /// and a tested type
-  TypeCast({required this.source, required this.target});
+  TypeCast({
+    required this.source,
+    required this.target,
+    this.isReversed = false,
+  });
 
   /// Returns the first type from source's supertypes
   /// which is corresponding to target or null
@@ -40,7 +47,7 @@ class TypeCast {
 
   /// Checks for nullable type casts
   /// Only one case `Type? is Type` always valid assertion case.
-  bool isNullableCompatibility() {
+  bool get isNullableCompatibility {
     final isObjectTypeNullable =
         source.nullabilitySuffix != NullabilitySuffix.none;
     final isCastedTypeNullable =
@@ -52,12 +59,9 @@ class TypeCast {
   /// Checks that type checking is unnecessary
   /// [source] is the source expression type
   /// [target] is the type against which the expression type is compared
-  /// [isReversed] true for opposite comparison, i.e 'is!'
   /// and false for positive comparison, i.e. 'is', 'as' or 'whereType'
-  bool isUnnecessaryTypeCheck({
-    bool isReversed = false,
-  }) {
-    if (isNullableCompatibility()) {
+  bool get isUnnecessaryTypeCheck {
+    if (isNullableCompatibility) {
       return false;
     }
 
@@ -70,7 +74,7 @@ class TypeCast {
       source: objectCastedType,
       target: target,
     );
-    if (!objectTypeCast.areGenericsWithSameTypeArgs()) {
+    if (!objectTypeCast.areGenericsWithSameTypeArgs) {
       return false;
     }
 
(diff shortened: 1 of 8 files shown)


lint_test/avoid_unnecessary_type_casts_test.dart
```diff
@@ -0,0 +1,18 @@
...
+  final testList = [1.0, 2.0, 3.0];
+
+  // to check quick-fix => testList
+  // expect_lint: avoid-unnecessary-type-casts
+  final result = testList as List<double>;
+
+  final double? nullableD = 2.0;
+  // casting `Type? is Type` is allowed
+  final castedD = nullableD as double;
+}
```
solid-yuriiprykhodko, line 18: Let's try to add more test cases.

For one, Dart Code Checker often stumbled on Map handling.

```
const m = {
  'a' : 'b'
 }

final mVal = m['a'] as String?;
```

I'd also check how this thing works with closures, passing arguments into a function etc.

lib/utils/typecast_utils.dart
```diff
@@ -36,4 +37,65 @@ class TypeCast {
...
+        when objectType is ParameterizedType && castedType is ParameterizedType
+    ) {
+      if (objectType.typeArguments.length != castedType.typeArguments.length) {
+        return false;
+      }
+
+      return IterableZip([objectType.typeArguments, castedType.typeArguments])
+        .every(
+          (e) => TypeCast(source: e[0], target: e[1]).isUnnecessaryTypeCheck(),
+        );
```
solid-yuriiprykhodko: These ops are lazy, so we might as well split them into granular, more readable pieces.

```suggestion
      return IterableZip([objectType.typeArguments, castedType.typeArguments])
        .map((e) => TypeCast(source: e[0], target: e[1]))
        .every((cast) => cast.isUnnecessaryTypeCheck());
```

lib/utils/typecast_utils.dart
```diff
@@ -36,4 +37,65 @@ class TypeCast {
...
+      source: objectCastedType,
+      target: target,
+    );
+
+    if (!objectTypeCast.areGenericsWithSameTypeArgs()) {
+      return false;
+    }
+
+    return !isReversed;
+  }
```
solid-yuriiprykhodko: I like the spacing, but when it's after every single statement -- it begins to lose meaning. Let's group minimally.
```suggestion
  bool isUnnecessaryTypeCheck({
        bool isReversed = false,
      }) {
    if (isNullableCompatibility()) return false;

    final objectCastedType = castTypeInHierarchy();
    if (objectCastedType == null) {
      return isReversed;
    }

    final objectTypeCast = TypeCast(
      source: objectCastedType,
      target: target,
    );
    if (!objectTypeCast.areGenericsWithSameTypeArgs()) {
      return false;
    }

    return !isReversed;
  }
```

solid-yuriiprykhodko (changes requested):

lib/utils/typecast_utils.dart
```diff
@@ -36,4 +37,63 @@ class TypeCast {
 
     return null;
   }
+
+  /// Checks for nullable type casts
+  /// Only one case `Type? is Type` always valid assertion case.
+  bool isNullableCompatibility() {
```
solid-yuriiprykhodko, line 43: ?
```suggestion
  bool get isNullableCompatibility {
```

lib/utils/typecast_utils.dart
```diff
@@ -36,4 +37,63 @@ class TypeCast {
...
+  }
+
+  /// Checks that type checking is unnecessary
+  /// [source] is the source expression type
+  /// [target] is the type against which the expression type is compared
+  /// [isReversed] true for opposite comparison, i.e 'is!'
+  /// and false for positive comparison, i.e. 'is', 'as' or 'whereType'
+  bool isUnnecessaryTypeCheck({
+    bool isReversed = false,
+  }) {
```
solid-yuriiprykhodko, line 59: isReversed could be a class member?
```suggestion
  bool get isUnnecessaryTypeCheck {
```

solid-yuriiprykhodko (changes requested):

lib/utils/typecast_utils.dart
DenisBogatirov: It can be, for the expense of creating on additional instance of TypeCast in avoid_unnecessary_type_assertions_rule. Should be fine

solid-yuriiprykhodko (approved):
Looking good!

PR #49, opened 2023-09-13
Add avoid unused parameters rule

DenisBogatirov:


lint_test/avoid_unused_parameters_test.dart
```diff
@@ -0,0 +1,46 @@
+// ignore_for_file: prefer_const_declarations
+// ignore_for_file: unnecessary_nullable_for_final_variable_declarations
+// ignore_for_file: unused_local_variable
+
+/// Check the `avoid-unused-parameters` rule
```
yurii-prykhodko-solid, line 5: Anonymous callbacks, closures?
Typedefs?
Underscore handling?
Constructors/factories?
Static methods?

lib/utils/parameter_utils.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:analyzer/dart/ast/ast.dart';
+
+/// Checks if parameter has underscores in its name
+bool hasNoUnderscoresInName(FormalParameter parameter) =>
+    parameter.name != null &&
+    parameter.name!.lexeme.replaceAll('_', '').isNotEmpty;
```
yurii-prykhodko-solid, line 6: Let's avoid `!`.
```suggestion
/// Checks if parameter has underscores in its name
bool hasNoUnderscoresInName(FormalParameter parameter) {
    final paramName = parameter.name
    if (paramName == null) return false; // That's what we would get from your version, but it seems odd. Should it be `true`?
    return paramName.lexeme.replaceAll('_', '').isNotEmpty;
    }
```

yurii-prykhodko-solid: Also, positive checks are usually easier to reason about. We can always negate it later, while negating an already negated function to create a positive check is weird to read.
hasNoUnderscoresInName -> hasUnderscoresInName

yurii-prykhodko-solid (changes requested):

lib/utils/parameter_utils.dart
yurii-prykhodko-solid: Wait, why isn't this a `String.contains()` check for that matter? 
It really checks whether the param consists _entirely_ of underscores.

DenisBogatirov: Function originates from DCM, and it's name gave me confusion it actually checks if param name consists only from underscores. Will rename and implement changes

DenisBogatirov: `String.contains('_')` will return true for '_param' and 'par_am', so we need to either count number of underscores and check it with string length or replace all underscores and look is something left or not. I've choose the latter.

lib/lints/avoid_unused_parameters/avoid_unused_parameters_rule.dart
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    context.registry.addClassDeclaration((node) {
+      final visitor = AvoidUnusedParametersVisitor();
+      visitor.visitClassDeclaration(node);
+
+      for (final element in visitor.unusedParameters) {
+        reporter.reportErrorForNode(code, element);
+      }
+    });
```
yurii-prykhodko-solid, line 61: Share the visitor and result?
```suggestion
    final visitor = AvoidUnusedParametersVisitor();
    context.registry.addMethodDeclaration(visitor.visitMethodDeclaration);
    context.registry.addFunctionDeclaration(visitor.visitFunctionDeclaration);
    context.registry.addClassDeclaration(visitor.visitClassDeclaration);
    
    for (final element in visitor.unusedParameters) {
      reporter.reportErrorForNode(code, element);
    }
```

yurii-prykhodko-solid (changes requested):

lib/lints/avoid_unused_parameters/avoid_unused_parameters_visitor.dart
```diff
@@ -119,16 +119,17 @@ class AvoidUnusedParametersVisitor extends RecursiveAstVisitor<void> {
...
       if (parameter is DefaultFormalParameter) {
-        /// Variables as 'Foo({super.param})' is being reported
-        /// as [DefaultFormalParameter] instead of [SuperFormalParameter]
-        /// it seems to be an issue in DartSDK
-        isSuperFormalParameter = parameter.toSource().startsWith('super.');
+        /// Variables as 'Foo({super.param})' or 'Foo({this.param})'
+        /// is being reported as [DefaultFormalParameter] instead
+        /// of [SuperFormalParameter] it seems to be an issue in DartSDK
+        isFieldFormalParameter = parameter.toSource().contains('this.');
+        isSuperFormalParameter = parameter.toSource().contains('super.');
```
yurii-prykhodko-solid, line 132: The workaround is nice and straightforward, the comment helps understand why we're doing this. Me likey!

lib/lints/avoid_unused_parameters/avoid_unused_parameters_rule.dart
DenisBogatirov: I saw a same node being reported multiple times when sharing the visitor. Will test it, and have idea how to do it differently if it not works

DenisBogatirov: for cycle should be inside of an callback, because `addSomething` is asynchronous. Did it through `addCompilationUnit` which is whole file and let visitor do its job

lint_test/avoid_unused_parameters_test.dart
```diff
@@ -0,0 +1,109 @@
...
+}
+
+class SomeOtherAnotherClass {
+  void method(String s) {
+    print(s);
+    return;
+  }
+
+  // expect_lint: avoid-unused-parameters
+  void anonymousCallback(Function(int a) cb) {}
```
yurii-prykhodko-solid, line 58:
```suggestion
  // expect_lint: avoid-unused-parameters
  void anonymousCallback(Function(int a) cb) {}
  void callbackUser() {
     // expect_lint: avoid-unused-parameters
    anonymousCallback((int a) {});
  }
```

yurii-prykhodko-solid (changes requested):
A couple questions on behavior.

lint_test/avoid_unused_parameters_test.dart
```diff
@@ -0,0 +1,109 @@
...
+void closure(int a) {
+  void internal(int a) {
+    print(a);
+  }
+}
+
+typedef MaxFun = int Function(int a, int b);
+
+// Allowed same way as override
+final MaxFun maxFunInstance = (int a, int b) => 1;
```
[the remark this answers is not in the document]
yurii-prykhodko-solid: @DenisBogatirov What's your take on this?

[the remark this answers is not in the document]
DenisBogatirov: Didn't quite get the question.

Type is `MaxFun` so we are obliged to provide given set of params, if we remove one then there would be a type mismatch. That would work same way with with name params because if we remove `namedParam` it wouldn't be a defined type.

Or do you want to create another test case for named params with typedef?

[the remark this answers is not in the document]
yurii-prykhodko-solid: With positional params, we have the freedom to create a placeholder arg with the correct type. Just need to match the signature.
Here's how I see it:

```dart
// expect_lint: avoid-unused-parameters
final MaxFun maxFunLint = (int a, int b) => 1;
  
final MaxFun maxFunOk = (int _, int __) => 1;
```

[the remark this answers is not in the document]
DenisBogatirov: Original rule does not care about parameter name as long as function or method is implemented.

Should I implement this logic? It will affect `@override` as well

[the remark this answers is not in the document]
yurii-prykhodko-solid: Hm, is there no way to know that we're operating on a typedef instead of a class method override?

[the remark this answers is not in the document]
DenisBogatirov: At least at this time I couldn't find any :(

lint_test/avoid_unused_parameters_test.dart
DenisBogatirov: Same issue as with [this comment](https://github.com/solid-software/solid_lints/pull/49#pullrequestreview-1628707372)

lint_test/avoid_unused_parameters_test.dart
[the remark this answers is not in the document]
yurii-prykhodko-solid: Well, in that case let's document this in an issue and get back to it later on.
You've done a great job expanding the rule to constructors already.

yurii-prykhodko-solid (approved):
LGTM!

PR #51, opened 2023-09-14
Feature/member ordering

DenisBogatirov:
Added member-ordering rule and implemented tests


lint_test/member_ordering_test.dart
```diff
@@ -0,0 +1,64 @@
...
+}
+
+class AWidget extends StatefulWidget {
+  const AWidget({super.key});
+
+  @override
+  State<AWidget> createState() => _AWidgetState();
+}
+
+class _AWidgetState extends State<AWidget> {
```
yurii-prykhodko-solid, line 35: Needs more fleshed-out testcases. Private fields, private methods, static members.
It would be much better to have _all_ that we enforce here.

yurii-prykhodko-solid (changes requested):

lint_test/member_ordering_test.dart
DenisBogatirov: Added more. But technically we can't test all because it has named-method pattern so it's possible to have config like
```
-a-method
-b-method
-c-method
-d-method
...
```

yurii-prykhodko-solid: I agree that testing all is not viable, but I would still ask for more.

Let's structure the tests a bit. Here's what I think will be helpful:

1. One all-positive case where we should see no lints at all.
2. One negative case with all ticks triggered.
3. One semi-negative case that would check that we can detect one isolated lint without triggering anything unrelated.

I'd also create two separate sets of tests for regular classes and for widgets, so there'll be 6 classes.

DenisBogatirov: Done

yurii-prykhodko-solid: I also see some code that would enfoce alphabetic order and alphabetic-by-type order. I don't see any tests for that -- just a couple basic ones will suffice.

DenisBogatirov: I can't test them both because you can pass only one of them at `analysis_options.yaml` will add for alphabetical.

yurii-prykhodko-solid: If we can't test it -- then it doesn't work. Let's try to find a way.
One option would be to create a subdir with its own `analysis_options.yaml` file -- but I'm not sure custom_lint handles such nesting well. Give it a test.
We can also create a completely separate package with its own config, with no nesting. A bit ugly, but at least we'll have it tested.
I don't think this will be the last rule with potentially conflicting configs.

yurii-prykhodko-solid (approved):
LGTM!

PR #62, opened 2023-09-20
Add prefer match file name

DenisBogatirov:


lint_test/prefer_match_file_name_test.dart
```diff
@@ -0,0 +1,17 @@
+// ignore_for_file: unused_element, unused_field
+
+/// Check the `prefer-match-file-name` rule
+class _AnotherPrivateClass {}
+
+class PreferMatchFileNameTest {}
```
yurii-prykhodko-solid, line 6: Should be a lint here.
```suggestion
// expect_lint: prefer-match-file-name
class PreferMatchFileNameTest {}
```

DenisBogatirov: It matches file name. 
Original rule from DCM ignores private declarations and checks only first public declaration

yurii-prykhodko-solid: Ah, my bad.

lint_test/prefer_match_file_name_test.dart
```diff
@@ -0,0 +1,18 @@
+// ignore_for_file: unused_element, unused_field
+
+/// Check the `prefer-match-file-name` rule
+class _AnotherPrivateClass {}
+
+// expect_lint: prefer-match-file-name
```
yurii-prykhodko-solid, line 6:
```suggestion
```

yurii-prykhodko-solid (approved):
LGTM

PR #67, opened 2023-11-10
Fix avoid_unused_parameters false reporting

### LINKED ISSUE #64: avoid-unused-parameters false-reports when parameter is used in constructor
image

solid-vovabeloded:
Closes #64


yurii-prykhodko-solid (approved):
Looks good!

PR #78, opened 2023-11-20
Set default severity level to warning

solid-vovabeloded:
Changes the default severity level of custom lints to warning.


illia-romanenko (approved):
LGTM

PR #84, opened 2023-11-22
Simpler member ordering

yurii-prykhodko-solid:


lib/analysis_options.yaml
```diff
@@ -63,23 +63,11 @@ custom_lint:
...
-        - static_fields
-        - static_methods
-        - public_fields
-        - private_fields
-        - constructors
-        - public_methods
-        - private_methods
         - init_state_method
         - build_method
         - did_change_dependencies_method
```
illia-romanenko, line 73: we also had deactivate in the list, should we add it here?

illia-romanenko (approved):
Looks good - one minor thing added let me know if we don't need it for some reason.

PR #85, opened 2023-11-22
Fix function_lines_of_code

yurii-prykhodko-solid:
Fixes `all` declarations being line-counted and linted.


illia-romanenko (approved):
LGTM

PR #89, opened 2023-11-28
New Lint: avoid_using_api

### LINKED ISSUE #81: Avoid Using API Lint Rule
I've been working on a custom lint rule for banning external code by the identifier. Is this something you would want PR'd to this project?

I started working on it before custom_lint exposed the lint options, so I have some refactoring to do. I would rather add contribute it here since it is already setup for it.

getBoolean:
Closes #81

This lint has a lot of edge cases. Tests are included to verify basic functionality.

## Packages Added:

* [glob](https://pub.dev/packages/glob): Determines which files an entry should apply to (better than Regex for paths)
* [yaml](https://pub.dev/packages/yaml): To convert `YamlMap` to `Map` for the list of entries

## Usage

External code can be "deprecated" when there is a better option available:

```yaml
custom_lint:
  rules:
    - avoid_using_api:
      severity: info
      entries:
        - class_name: Future
          identifier: wait
          source: dart:async
          reason: "Future.wait should be avoided because it loses type safety for the results. Use a Record's `wait` method instead."
          severity: warning
```

Result:

```dart
void main() async {
  await Future.wait([...]); // LINT
  await (...,).wait; // OK
}
```

### Advanced Usage

Each entry also has `includes` and `excludes` parameters. These paths utilize [Glob](https://pub.dev/packages/glob) patterns to determine if a lint entry should be applied to a file.

For example, a lint to prevent usage of a domain-only package outside of the domain folder:

```yaml
custom_lint:
  rules:
    - avoid_using_api:
      severity: info
      entries:
        - source: package:modddels
          excludes:
            - "**/domain/**.dart"
          reason: "MODDDELS is only intended to be used in the domain layer."
```


lib/lints/banned_external_code/banned_external_code_rule.dart
```diff
@@ -0,0 +1,315 @@
...
+  final ErrorReporter reporter;
+  final CustomLintContext context;
+  final RuleConfig<BannedExternalCodeParameters?> config;
+
+  bool _matchesSource(String? parentSourcePath, String sourceToMatch) {
+    if (parentSourcePath == null) {
+      return false;
+    }
+
+    final libraryPathParts = parentSourcePath.split('/');
```
yurii-prykhodko-solid, line 130: Use Platform.pathSeparator for Windows compat?

lib/utils/parameter_utils.dart
```diff
@@ -8,3 +13,133 @@ bool nameConsistsOfUnderscoresOnly(FormalParameter parameter) {
...
+}
+
+/// Normalizes the path using posix style and
+/// replaces backslashes with forward slashes
+String normalizePath(String path, [String? root]) {
+  // Remove drive letter from path
+  final String subpath = _removeDriveLetter(path);
+
+  return p.posix.relative(subpath.replaceAll(r'\', '/'), from: root);
+}
```
solid-vovabeloded, line 130: Please, use `Uri.normalizePath` instead.

lint_test/banned_external_code/id_ban/lib/id_ban.dart
```diff
@@ -0,0 +1,30 @@
...
+  print('Hello World');
+}
+
+class BannedCodeUsage {
+  BannedCodeUsage();
+
+  void test() {
+    print('Hello World');
+  }
+
```
yurii-prykhodko-solid, line 26:
```suggestion
  void test() {
    print('Hello World');
  }
  void useTest() {
    test(); // implicit `this`, no lint expected
  }
```

lib/lints/banned_external_code/banned_external_code_rule.dart
```diff
@@ -0,0 +1,315 @@
...
+  ) {
+    banIdFromSource(entryCode, id);
+  }
+
+  void banClass(
+    LintCode entryCode,
+    String className,
+  ) {
+    banClassFromSource(entryCode, className);
+  }
```
yurii-prykhodko-solid, line 161: Not sure these trampoline methods are useful. Maybe we should call the underlying function directly?

example/analysis_options.yaml
```diff
@@ -19,3 +19,11 @@ custom_lint:
     - avoid_unnecessary_setstate
     - double_literal_format
     - avoid_unnecessary_type_assertions
+    - banned_external_code:
```
solid-vovabeloded, line 22: What do you think about renaming this lint rule to `avoid_using_api`?

lib/utils/parameter_utils.dart
```diff
@@ -8,3 +13,133 @@ bool nameConsistsOfUnderscoresOnly(FormalParameter parameter) {
...
+
+bool _isFileIncluded(List<String> includes, String path) {
+  final hasMatch =
+      includes.map(Glob.new).toList().any((glob) => glob.matches(path));
+  return hasMatch;
+}
+
+bool _isFileNotIncluded(List<String> includes, String path) {
+  return !_isFileIncluded(includes, path);
+}
```
yurii-prykhodko-solid, line 121: Could we dedupe a bit?
```suggestion
bool _matchesAnyGlob(List<String> globsList, String path) {
  final hasMatch =
      globsList.map(Glob.new).toList().any((glob) => glob.matches(path));
  return hasMatch;
}

bool _doesNotMatchGlobs(List<String> globList, String path) {
  return !_matchesAnyGlob(globList, path);
}
```

lib/lints/banned_external_code/models/banned_external_code_entry_parameters.dart
```diff
@@ -0,0 +1,78 @@
...
+/// ```
+class BannedExternalCodeEntryParameters {
+  /// Variable/method name
+  final String? id;
+
+  /// Name of the class containing the variable/method
+  final String? className;
+
+  /// Package (e.g., dart:async or package:example)
+  final String? source;
```
yurii-prykhodko-solid, line 36: IMO `source` should be required, and limited to external packages.

Whatever code is in the current package -- we can safely assume we have control of, and there already is a `@Deprecated` mechanism in place for cases where we want to phase out the usage of some internal API.

I'd rather not introduce an alternative way to do the same thing.

lint_test/banned_external_code/id_ban/analysis_options.yaml
```diff
@@ -0,0 +1,11 @@
+analyzer:
+  plugins:
+    - ../../custom_lint
```
yurii-prykhodko-solid, line 3: Doesn't seem to work correctly otherwise.
```suggestion
    - custom_lint
```

example/analysis_options.yaml
yurii-prykhodko-solid: `deprecate_external_dependency`?

lint_test/banned_external_code/id_ban/analysis_options.yaml
getBoolean: I'm pretty sure the issue is from [the analysis_options.yaml](https://github.com/invertase/dart_custom_lint/issues/148) file in the `lib/` folder. After deleting it the lints show up in the IDE

lib/lints/banned_external_code/banned_external_code_rule.dart
getBoolean: Yeah I agree is it a little weird, it's a holdover from before I refactored it to remove the duplicate code. I decided to keep it for the sake of consistency/readability.

If we make source required, these two can be removed entirely.

example/analysis_options.yaml
```diff
@@ -19,3 +19,11 @@ custom_lint:
...
+      severity: info
+      entries:
+        - class_name: Future
+          id: wait
```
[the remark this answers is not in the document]
getBoolean: I think that would imply it has to be a member of a class, when it can also be the name of a method/variable in the global namespace of a source.

lib/lints/banned_external_code/banned_external_code_rule.dart
```diff
@@ -0,0 +1,315 @@
...
+
+    final matchesLibraryName = libraryName == sourceLibraryName;
+    final shouldLintFileFromSource = sourceLibraryPathParts.length > 1;
+    if (shouldLintFileFromSource) {
+      final sourceFile = (sourceLibraryPathParts..removeAt(0)).join('/');
+      final libraryFile = (libraryPathParts..removeAt(0)).join('/');
+
+      final matchesFile = p.equals(sourceFile, libraryFile);
+      return matchesLibraryName && matchesFile;
+    }
```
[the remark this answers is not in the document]
getBoolean: Is this what you had in mind?

```dart
    final Uri parentUri = p.toUri(parentSourcePath);
    final Uri sourceUri = p.toUri(sourceToMatch);

    final libraryName = parentUri.pathSegments.first;
    final sourceLibraryName = sourceUri.pathSegments.first;

    final matchesLibraryName = libraryName == sourceLibraryName;
    final shouldLintFileFromSource = sourceUri.pathSegments.length > 1;
    if (shouldLintFileFromSource) {
      final sourceFile = sourceUri.pathSegments.sublist(1).join('/');
      final libraryFile = parentUri.pathSegments.sublist(1).join('/');

      final matchesFile = p.equals(sourceFile, libraryFile);
      return matchesLibraryName && matchesFile;
    }
```

lib/lints/banned_external_code/banned_external_code_rule.dart
```diff
@@ -0,0 +1,315 @@
...
+  ]) {
+    context.registry.addSimpleIdentifier((node) {
+      final parentSourceName = node.sourceUrl;
+      if (source != null && !_matchesSource(parentSourceName, source)) {
+        return;
+      }
+
+      final parent = node.parent;
+      final entityBeforeNode =
+          parent!.childEntities.firstWhere((element) => element != node);
```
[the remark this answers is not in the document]
getBoolean: It was a bandaid fix, from before I intended to publish this. I've pushed a fix to the real issue, which was `ConstructorDeclaration` being linted (it should not lint declarations)

lib/utils/node_utils.dart
```diff
@@ -22,3 +22,23 @@ String humanReadableNodeType(AstNode? node) {
...
+
+/// Extension to get library of the node
+extension NamedTypeParentSourceExtension on NamedType {
+  /// Returns library of the node
+  String? get sourceUrl {
+    final parentSource = element?.librarySource;
+    final parentSourceName = parentSource?.uri.toString();
+    return parentSourceName;
+  }
+}
```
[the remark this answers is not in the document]
getBoolean: I've moved them to a new file in `banned_external_code` folder.

lib/utils/parameter_utils.dart
```diff
@@ -8,3 +13,133 @@ bool nameConsistsOfUnderscoresOnly(FormalParameter parameter) {
...
+        name: name ?? this.name,
+        problemMessage: problemMessage ?? this.problemMessage,
+        correctionMessage: correctionMessage ?? this.correctionMessage,
+        uniqueName: uniqueName ?? this.uniqueName,
+        url: url ?? this.url,
+        errorSeverity: errorSeverity ?? this.errorSeverity,
+      );
+}
+
+/// Source: https://github.com/epam-cross-platform-lab/swagger-dart-code-generator/blob/master/lib/src/extensions/yaml_extensions.dart
```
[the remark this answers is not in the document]
getBoolean: Sorry, this attribution is probably incorrect.

Upon further research, I think the correct source of this code is this [stackoverflow](https://stackoverflow.com/questions/23143653/how-to-load-yaml-file-to-the-map-in-dart) post, although it is not where I originally got it from. How should this be attributed?

A search for [YamlMapConverter on GitHub](https://github.com/search?q=YamlMapConverter&type=code&p=1) reveals many entries of this exact same code block with no attribution, so I was not sure what to put.

lib/utils/parameter_utils.dart
```diff
@@ -8,3 +13,133 @@ bool nameConsistsOfUnderscoresOnly(FormalParameter parameter) {
...
+  }
+
+  if (isExcludeOnly) {
+    return _isFileExcluded(includeGlobs, pathNormalized);
+  }
+
+  final isFileNotIncluded = _isFileNotIncluded(includeGlobs, pathNormalized);
+  final isFileExcluded = _isFileExcluded(excludeGlobs, pathNormalized);
+  return isFileNotIncluded || isFileExcluded;
+}
```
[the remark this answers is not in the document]
getBoolean: Removing `if (includeGlobs.isEmpty && excludeGlobs.isEmpty)` changes the behavior of the function. I am not sure about `isIncludeOnly` or `isExcludeOnly` yet, I need to setup a test for them.

This function is a little unintuitive, so I'll walk through it and convert the terminology to positives (this is mostly for my own understanding, it is a bit confusing)

* Line 95: If only include globs are given, only analyze the file if the path matches one of the include globs
* Line 99: If only exclude globs are given, only analyze the file if the path does not match any of the exclude globs
* Line 102-104: only analyze the file if the path matches one of the include globs and does not match any of the exclude globs

Maybe the function `shouldSkipFile` should be flipped and renamed to `shouldAnalyzeFile`?

lib/utils/parameter_utils.dart
getBoolean: Alright, I've also renamed it so it is more accurate to what it returns. For context, the globs fail when the drive letter is in the path so using relative is needed, and it doesn't make sense to match against folders above the project.
```dart
/// Converts path to relative using posix style and
/// replaces backslashes with forward slashes
String relativePath(String path, [String? root]) {
  final uriNormlizedPath = p.toUri(path).normalizePath().path;
  final uriNormlizedRoot =
      root != null ? p.toUri(root).normalizePath().path : null;

  final relative = p.posix.relative(uriNormlizedPath, from: uriNormlizedRoot);
  return relative;
}
```

example/analysis_options.yaml
getBoolean: I'll leave the final decision up to you and your team

lib/utils/parameter_utils.dart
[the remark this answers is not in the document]
getBoolean: Refactored version:

```dart
bool shouldSkipFile({
  required List<String> includeGlobs,
  required List<String> excludeGlobs,
  required String path,
  String? rootPath,
}) {
  final relative = relativePath(path, rootPath);
  final shouldAnalyzeFile =
      (includeGlobs.isEmpty || _matchesAnyGlob(includeGlobs, relative)) &&
          (excludeGlobs.isEmpty || _doesNotMatchGlobs(excludeGlobs, relative));
  return !shouldAnalyzeFile;
}

bool _matchesAnyGlob(List<String> globsList, String path) {
  final hasMatch =
      globsList.map(Glob.new).toList().any((glob) => glob.matches(path));
  return hasMatch;
}

bool _doesNotMatchGlobs(List<String> globList, String path) {
  return !_matchesAnyGlob(globList, path);
}
```

yurii-prykhodko-solid:
@getBoolean Apologies for a tardy response -- I see you've made quite a bit of progress!

> 1. Rename parameter `id` to something that represents methods and variables, not sure what to name it.

Maybe `pattern`? You are using glob matching under the hood, seems consistent.

> 2. Make `source` required. What should it do if `source` is not provided?

I'd say just ignore it, and provide ample warning in rule documentation & usage examples.

> 3. Rename lint. I'll add `deprecate_external_api` as an additional choice

Oh this one sparked quite a discussion :D

The consensus was to use the `avoid_using_api` name.

lib/lints/banned_external_code/models/banned_external_code_entry_parameters.dart
yurii-prykhodko-solid: > I agree, though we can't stop users from specifying the source as package:my_app.

Well, if the users want to use a hacky loophole -- who are we to stop their creativity :)

getBoolean:
@yurii-prykhodko-solid glob pattern matching is only used for includes/excludes filepaths, not the id/class name. Is that something we want to change?

yurii-prykhodko-solid:
@getBoolean Ah, confused the two. Then `pattern` is no good.

`lexeme`? May be too broad in meaning.
`symbol`? Not really a common term in context of Dart lang.

And both of the above seem just as confusing as `id` to me. So I'd say we can leave it as is.

@solid-vovabeloded maybe you have some ideas?

getBoolean:
Where should I put the lint documentation?

yurii-prykhodko-solid:
@getBoolean 
> * Rename parameter `id` to something that represents methods and variables, not sure what to name it.

Discussed this internally, the consensus was to use `identifier`, with no abbreviation. "Identifier" is the term Dart team uses for everything named -- functions, variables, classnames, which seems consistent with what the lint is expecting.

> Where should I put the lint documentation?

That is an interesting question -- so far we've been maintaining this project as a drop-in solution, keeping configuration as minimal as it's possible, and therefore with little library user facing documentation.

I'd say we can leave this as an issue on this repo until we come up with a general approach of documenting our lints.

getBoolean:
I think that's everything!

I do think documentation is needed, it will be very useful to know what all the lints are and what triggers them. Documentation for lints options can be included alongside that.

yurii-prykhodko-solid (commented):
Looks good overall, great piece of work, @getBoolean!

Let's make the CI happy and we can merge.

getBoolean:
I hope that is everything.

yurii-prykhodko-solid (approved):
Outstanding piece of work, @getBoolean!

illia-romanenko:
Nice! Our CI still fails - it feels like we need to fix those issues before merging?

getBoolean:
Not sure why its failing, I'll look into it when I get home from vacation in a week.

getBoolean:
I've confirmed it working correctly on Windows and MacOS. It might be an issue somewhere in `AvoidUsingApiLinter._matchesSource` on Linux or CI causing the source check to fail.

getBoolean:
Good thing I setup Windows/Linux dual boot already, made tracking down the issue a lot easier.

The issue was that `node.sourceUrl` would return the filepath on Linux for project files, while it was always the package path on Windows/MacOS. In the documentation, maybe we should specifically state that doing `source: package:myapp` will not work on Linux and should be avoided.

I also found two edge cases while debugging and fixed them

- lint for method usage on variable of a banned class type (source was previously ignored for it)
- lint for variable declaration of a banned class type if the type was inferred.

yurii-prykhodko-solid:
> In the documentation, maybe we should specifically state that doing `source: package:myapp` will not work on Linux and should be avoided.

Which makes sense -- that's what 'Deprecated' is for.

Let's add this as a note in code documentation, I think the `AvoidUsingApiParameters` docs are the right place.

yurii-prykhodko-solid:
> Let's add this as a note in code documentation, I think the AvoidUsingApiEntryParameters docs are the right place.

Done.

@getBoolean Let me know if you're happy with my changes, then we can finish this up.

yurii-prykhodko-solid:
@getBoolean Looks like there's some more issues -- CI still in the red.

getBoolean:
Before we merge, I want to investigate making `_matchesSource` take a non nullable parent source url, it was hiding a lint issue before.

@yurii-prykhodko-solid it looks like pub get needs to be run on those test packages in CI, is that possible?

yurii-prykhodko-solid:
> @yurii-prykhodko-solid it looks like pub get needs to be run on those test packages in CI, is that possible?

Sure -- have a stab at our CI file. A `find -> exec` or a `find -> for` should do the trick.

getBoolean:
Thanks, I think that is everything now

PR #97, opened 2024-01-15
Documentation parsing and Markdown doc generation

solid-daniilmarchenko:
Documentation parser for the solid lints

[Example of output](https://github.com/solid-software/solid_lints/blob/documentation-parser/DOCUMENTATION.md)


lib/utils/docs_parser/output_formatters/base_formatter.dart
```diff
@@ -0,0 +1,23 @@
...
+    for (final rule in rules) {
+      formattedResult.writeln(formatRuleDoc(rule));
+    }
+
+    return formattedResult.toString();
+  }
+
+  /// Get formatted string for the single given [rule]
+  String formatRuleDoc(RuleDoc rule);
+}
```
yurii-prykhodko-solid, line 23: Think about how this may support outputting a file per rule. For now the base class does little to support it.

Perhaps we can keep it as simple as possible. Let the concrete implementation determine what the outputs are:

```dart
interface class RulesDocumentationFormatter<T> {
  T format(List<RuleDoc> docs);
}
```

lib/utils/docs_parser/parsers/parameters_parser.dart
```diff
@@ -0,0 +1,88 @@
...
+    final parameterDocs = <ParameterDoc>[];
+    for (final declaration in ast.unit.declarations) {
+      if (declaration is ClassDeclaration) {
+        parameterDocs.addAll(_parseParametersDocs(declaration));
+        if (parameterDocs.isNotEmpty) break;
+      }
+    }
+
+    return parameterDocs;
+  }
```
yurii-prykhodko-solid, line 41: Same as the first comment. May need `package:collection` if it's not there yet: 
```suggestion
    final parameterDocs = ast.unit.declarations
      .whereType<ClassDeclaration>()
      .map(_parseParametersDocs)
      .firstWhereOrNull((docs) => docs.isNotEmpty)
      .expand((e) => e);

    return parameterDocs;
  }
```

lib/utils/docs_parser/parsers/parameters_parser.dart
```diff
@@ -0,0 +1,88 @@
...
+        final name = ParserUtils.camelCaseToSnakeCase(variableName);
+        final type = member.fields.type.toString();
+        final doc = ParserUtils.formatDocumentationComment(
+          member.documentationComment,
+        );
+
+        if (doc == null) {
+          throw 'Documentation is not specified for class: '
+              '${declaration.name.stringValue}';
+        }
```
yurii-prykhodko-solid, line 61: Log a warning, don't throw.

yurii-prykhodko-solid (changes requested):
Let's make the output formatter base class a bit more lax in its requirements, everything else looks fine.

yurii-prykhodko-solid:
@solid-daniilmarchenko maybe also attach an example of how the generated doc looks like?

solid-daniilmarchenko:
> @solid-daniilmarchenko maybe also attach an example of how the generated doc looks like?

Changed the PR description and included link to the output file

DOCUMENTATION.md
```diff
@@ -0,0 +1,165 @@
...
+A `prefer_match_file_name` rule which warns about
+ mismatch between file name and declared element inside
+
+
+### prefer_conditional_expressions
+A `prefer_conditional_expressions` rule which warns about
+ simple if statements that can be replaced with conditional expressions
+#### Parameters:
+- **ignore_nested** (_bool_)  
+  Should rule ignore nested if statements
```
yurii-prykhodko-solid, line 11: This looks a bit better, IMHO:

```suggestion
## prefer_match_file_name
A `prefer_match_file_name` rule which warns about
 mismatch between file name and declared element inside


## prefer_conditional_expressions
A `prefer_conditional_expressions` rule which warns about
 simple if statements that can be replaced with conditional expressions
### Parameters:
- **ignore_nested** (_bool_)  
  Should rule ignore nested if statements
```

DOCUMENTATION.md
```diff
@@ -0,0 +1,165 @@
+### prefer_match_file_name
+A `prefer_match_file_name` rule which warns about
+ mismatch between file name and declared element inside
+
+
+### prefer_conditional_expressions
```
yurii-prykhodko-solid, line 6: Can we add a top-level header and a Table Of Contents?

yurii-prykhodko-solid (changes requested):

DOCUMENTATION.md
solid-daniilmarchenko: oh, yeah, we discussed table of contents and i totally forgot to implement it. question of one commit!

DOCUMENTATION.md
```diff
@@ -0,0 +1,211 @@
+# Solid Lints Documentation
+
+## Table of contents:
+
+1. [avoid_global_state](#avoid_global_state)
+1. [avoid_late_keyword](#avoid_late_keyword)
+1. [avoid_non_null_assertion](#avoid_non_null_assertion)
```
yurii-prykhodko-solid, line 7: Generate correct numbers here?

```suggestion
1. [avoid_global_state](#avoid_global_state)
2. [avoid_late_keyword](#avoid_late_keyword)
3. [avoid_non_null_assertion](#avoid_non_null_assertion)
```

yurii-prykhodko-solid (changes requested):

DOCUMENTATION.md
yurii-prykhodko-solid: Yeah, but having these numbers be readable in raw markdown would be nice.

solid-daniilmarchenko:
I've just found out that we don't have generated docs for `cyclomatic_complexity` and some other rules, just becouse they are called with `metric` suffix. I gonna fix that by adding `metric` suffix to the list of parsed ones

DOCUMENTATION.md
solid-daniilmarchenko: done

yurii-prykhodko-solid (approved):
LGTM!

PR #103, opened 2024-01-18
Docusaurus web docs

solid-daniilmarchenko:
Example of deployed instance: https://solid-daniilmarchenko.github.io/solid_lints/


doc/docusaurus/docs/Lints Documentation/avoid_late_keyword.md
```diff
@@ -0,0 +1,11 @@
+A `late` keyword rule which forbids using it to avoid runtime exceptions.
```
yurii-prykhodko-solid, line 1: Let's not keep generated files in VCS if we can help it.

lib/utils/docs_parser/parse_docusaurus_docs.dart
```diff
@@ -0,0 +1,15 @@
+import 'dart:io';
+
+import 'package:path/path.dart';
+import 'package:solid_lints/utils/docs_parser/output_formatters/docusaurus_formatter.dart';
+import 'package:solid_lints/utils/docs_parser/parsers/docs_parser.dart';
+
+void main() async => DocsParser(
```
yurii-prykhodko-solid, line 7: Let's move all executable scripts under `/bin`.

lib/utils/docs_parser/parse_docusaurus_docs.dart
```diff
@@ -0,0 +1,15 @@
...
+
+void main() async => DocsParser(
+      formatter: DocusaurusFormatter(),
+    ).parse(
+      Directory(
+        normalize(
+          join(Directory.current.path, 'lib', 'lints'),
+        ),
+      ),
+    );
```
yurii-prykhodko-solid, line 15: Let's add input-output parameters properly. For now we seem to be hardcoding a good lot, and it's really not obvious where things come from and where they go -- will make it tricky to figure out once we get back to this in some time.

Let's add some simple parameters that would make it apparent what the script takes in and what it outputs.

.github/workflows/gh-pages-test-deploy.yaml
```diff
@@ -0,0 +1,29 @@
+name: Test deployment
```
yurii-prykhodko-solid, line 1: Is this needed in `master`?

bin/parse_web_docs.dart
```diff
@@ -0,0 +1,73 @@
...
+      mandatory: true,
+    )
+    ..addOption(
+      'docs-dir',
+      abbr: 'o',
+      help: 'Parser output path. i.e "docusaurus/docs" directory',
+      defaultsTo: docusaurusDefaultPath,
+    )
+    ..addOption(
+      'output-dir',
```
yurii-prykhodko-solid, line 31:
```suggestion
      'output-dirname',
```

yurii-prykhodko-solid: Scratch it -- just remove the argument and make the path part of the `docs-dir` argument.

bin/parse_web_docs.dart
```diff
@@ -0,0 +1,73 @@
+import 'dart:io';
```
yurii-prykhodko-solid, line 1: rename to `generate_doc_templates.dart`

.github/workflows/gh-pages-test-deploy.yaml
solid-daniilmarchenko: yep, this checks if documentation could be successfully built

bin/parse_web_docs.dart
solid-daniilmarchenko: it's necessary for separating outer folder which contains `intro.md` file and the inner folder with lints documenation

solid-daniilmarchenko: but, after thinking a little, i think we could resolve path for `intro.md` file by just using parent directory of output dir. 
will change it tomorrow

PR #104, opened 2024-01-19
Update `analyzer` dependency

solid-daniilmarchenko:


solid-daniilmarchenko:
Is this behaviour expected?  
Seems like a bug.

Screenshot 2024-01-19 at 16 23 47


We have two linter warnings around one function: the one right before the 'void' keyword and second near brackets. 

Screenshot 2024-01-19 at 16 22 12

solid_lints.iml
```diff
@@ -9,6 +9,48 @@
       <excludeFolder url="file://$MODULE_DIR$/example/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/example/build" />
       <excludeFolder url="file://$MODULE_DIR$/example/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/lint_test/.dart_tool" />
```
yurii-prykhodko-solid, line 12: This file probably doesn't belong in the VCS?

yurii-prykhodko-solid:
> Is this behaviour expected?
@solid-daniilmarchenko yeah looks like a bug -- could be stemming from the fact that we're checking both function declaration and function body -- where one contains another, IIRC. Maybe just check function bodies? Give it a try.

solid-daniilmarchenko:
it looks like we really don't need `FunctionDeclaration` but removing it doesn't really fix all of the warning duplications.
i've tried to check other node types (including the `BlockFunctionBody`) that we could use instead of the current ones but it doesnt look to work fine anyway.

lib/lints/function_lines_of_code/function_lines_of_code_metric.dart
```diff
@@ -50,7 +50,6 @@ class FunctionLinesOfCodeMetric
 
     context.registry.addMethodDeclaration(checkNode);
     context.registry.addFunctionExpression(checkNode);
-    context.registry.addFunctionBody(checkNode);
```
yurii-prykhodko-solid: Seems to have done the trick with linting all we want with no duplicates

yurii-prykhodko-solid (approved):
LGTM!

PR #106, opened 2024-01-24
Web docs

solid-daniilmarchenko:


lib/utils/docs_parser/models/parameter_doc.dart
```diff
@@ -1,18 +1,18 @@
-/// Single parameter documentation
+/// Lint rule parameter documentation
```
yurii-prykhodko-solid, line 1: Let's move all the utils under `src`

doc/docusaurus/docusaurus.config.js
```diff
@@ -0,0 +1,105 @@
...
+  title: 'solid_lints',
+  tagline: 'solid_lints are cool',
+  url: 'https://solid-software.github.io/',
+  baseUrl: '/solid_lints/',
+  onBrokenLinks: 'throw',
+  onBrokenMarkdownLinks: 'warn',
+  favicon: 'img/favicon.png',
+  organizationName: 'solid_software', // Usually your GitHub org/user name.
+  projectName: 'solid_lints', // Usually your repo name.
+  deploymentBranch: 'web-docs-deployment',
```
solid-daniilmarchenko, line 18: before merge, we should be sure that this branch is exists and it is selected as default branch for github pages in settings.

bin/generate_web_docs_content.dart
```diff
@@ -0,0 +1,73 @@
...
+      abbr: 'o',
+      help: 'Parser output path. i.e "docusaurus/docs" directory',
+      defaultsTo: docusaurusDefaultPath,
+    )
+    ..addOption(
+      'output-dirname',
+      abbr: 'n',
+      help: 'Output directory name for generated lints documentation files',
+      defaultsTo: 'Lints Documentation',
+    )
```
yurii-prykhodko-solid, line 35: Remove this arg, make dirname part of the `docs-dir` path.

doc/docusaurus/docusaurus.config.js
```diff
@@ -0,0 +1,105 @@
+// @ts-check
```
yurii-prykhodko-solid, line 1: There used to be a bunch of docs pulled from `analysis-options.yaml`, with rationale on why we did or did not enable some of the standard rules. Can we restore it?

yurii-prykhodko-solid: Ah, nevermind -- looks like I never moved much of the docs into Markdown templates.

We can pull the comments from analysis-options.yaml in a follow-up.

solid-daniilmarchenko: i've restored all of the previous docs to the `/static/docs` directory, so we can select the ones that are useful

yurii-prykhodko-solid: Cool, thanks -- I'll give these a trim now.

doc/docusaurus/src/components/HomepageFeatures.tsx
```diff
@@ -11,7 +11,7 @@ type FeatureItem = {
 const FeatureList: FeatureItem[] = [
   {
     title: 'Write clean, maintainable code',
-    image: '/img/gears-gear-svgrepo-com.svg',
+    image: '/solid_lints/img/gears-gear-svgrepo-com.svg',
```
solid-daniilmarchenko, line 14: so any of the [documentation's suggested](https://docusaurus.io/docs/static-assets#referencing-your-static-asset) options haven't worked for me. i've decided to just hardcode url since it's not supposed to be changed frequently

doc/docusaurus/docusaurus.config.js
```diff
@@ -0,0 +1,106 @@
...
+
+const lightCodeTheme = require('prism-react-renderer').themes.github;
+const darkCodeTheme = require('prism-react-renderer').themes.dracula;
+
+/** @type {import('@docusaurus/types').Config} */
+const config = {
+  title: 'solid_lints',
+  tagline: 'An opinionated set of lint rules based on industry standards.',
+  url: 'https://solid-software.github.io/',
+  baseUrl: '/solid_lints/',
```
yurii-prykhodko-solid, line 12: We don't really need this, do we?
```suggestion
  baseUrl: '/',
```

solid-daniilmarchenko: Screenshot 2024-01-24 at 15 00 39

yurii-prykhodko-solid: This is odd -- I think I used just the slash when making changes recently

PR #107, opened 2024-01-24
Update dependencies

yurii-prykhodko-solid:


doc/docusaurus/.gitignore
```diff
@@ -18,3 +18,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+package-lock.json
```
solid-daniilmarchenko, line 22:
```suggestion

package-lock.json
yarn.lock
```
?

yurii-prykhodko-solid: Let's keep yarn lockfile. JIC we need to debug some dependency issues (sincerely hope not)

The reason for removing `package-lock.json` was that we should stick to a single dependency management tool, and since docusaurus recommends `yarn` -- let's just keep only that.

solid-daniilmarchenko (approved):

PR #109, opened 2024-01-25
Update README.md

solid-daniilmarchenko:


yurii-prykhodko-solid (changes requested):
This should probably be at the very top of the page for convenience.

We should probably lead directly to the docs overview ( https://solid-software.github.io/solid_lints/docs/intro ). 
Once again -- for convenience. Less clicks == happy user.

yurii-prykhodko-solid (approved):
Thanks, @solid-daniilmarchenko!

PR #118, opened 2024-01-30
Add args dependency; move internal tooling to tool dir; add documentation link and pub topics

yurii-prykhodko-solid:
```
Validating package... (4.8s)
Package has 0 warnings.
The server may enforce additional checks.
```


illia-romanenko (approved):
LGTM

PR #120, opened 2024-01-31
Redirect from index page to intro, add backup link for manual redirection

yurii-prykhodko-solid:
If redirection fails for any reason, show the following:

image
image


illia-romanenko (approved):
LGTM

PR #125, opened 2024-02-05
Bump custom_lint version to 0.6.0

solid-vovabeloded:
Bumps `custom_lint` package version to 0.6.0 and removes unnecessary nested packages from the `lint_test` directory.


illia-romanenko (approved):
LGTM

PR #141, opened 2024-03-13
 - Added `excludeNames` param for `function_lines_of_code` lint

DerhachevAndrii:
- Added `excludeNames` param for `function_lines_of_code` lint


lib/src/lints/function_lines_of_code/function_lines_of_code_metric.dart
```diff
@@ -57,6 +60,22 @@ class FunctionLinesOfCodeMetric
...
+      if (node is FunctionDeclaration) {
+        return node.name.lexeme;
+      } else if (node is MethodDeclaration) {
+        return node.name.lexeme;
+      } else if (node is FunctionExpression) {
+        return node.declaredElement?.name;
+      } else {
+        return null;
+      }
+    }();
```
solid-vovabeloded, line 73: Please, extract this into a separate private method or just inline it into current method. Do not use inline function declaration since it hurts readability.

lib/src/lints/function_lines_of_code/models/function_lines_of_code_parameters.dart
```diff
@@ -5,16 +5,22 @@ class FunctionLinesOfCodeParameters {
   /// exceeding this limit triggers a warning.
   final int maxLines;
 
+  /// Method names to be excluded from the rule check
```
solid-vovabeloded, line 8:
```suggestion
  /// Function names to be excluded from the rule check
```
Does it make sense to be consistent with all the rest documentation/naming?

solid-vovabeloded (changes requested):
Added a few suggestions here, but looks good overall.

lint_test/function_lines_of_code_test/function_lines_of_code_test.dart
```diff
@@ -0,0 +1,41 @@
...
+  // Excluded by excludeNames
+  int longMethodExcluded() {
+    var i = 0;
+    i++;
+    i++;
+    i++;
+    i++;
+    return i;
+  }
+}
```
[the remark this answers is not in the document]
DerhachevAndrii: done

solid-vovabeloded (approved):

PR #142, opened 2024-03-13
Excluded methods/functions named 'copyWith' from `number_of_parameter…

Viktor-Mudrytskyi:
…s` lint


yurii-prykhodko-solid:
We should do this using the common updated identifier exclusion mechanism.

PR #143, opened 2024-03-14
 - Fixed unexpected avoid_unnecessary_type_assertions

DerhachevAndrii:
- Fixed unexpected avoid_unnecessary_type_assertions

### DIFF
--- a/lib/src/utils/typecast_utils.dart
+++ b/lib/src/utils/typecast_utils.dart
@@ -67,7 +67,7 @@ class TypeCast {
 
     final objectCastedType = castTypeInHierarchy();
     if (objectCastedType == null) {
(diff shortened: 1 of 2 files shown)


lib/src/utils/typecast_utils.dart
```diff
@@ -67,7 +67,7 @@ class TypeCast {
 
     final objectCastedType = castTypeInHierarchy();
     if (objectCastedType == null) {
-      return isReversed;
+      return false;
```
solid-vovabeloded, line 70: With this fix, it won't lint the following situation: 
изображение

solid-vovabeloded (changes requested):

PR #154, opened 2024-03-28
Add a rule `prefer_guard_clause` for reversing nested if statements

velvitoff:


lib/src/lints/prefer_guard_clause/prefer_guard_clause_rule.dart
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @override
+  void run(
+    CustomLintResolver resolver,
+    ErrorReporter reporter,
+    CustomLintContext context,
+  ) {
+    context.registry.addDeclaration((node) {
+      if (node is! FunctionDeclaration && node is! MethodDeclaration) {
```
yurii-prykhodko-solid, line 61: Does it have to be like this maybe?
```suggestion
      if (node is! FunctionDeclaration || node is! MethodDeclaration) {
```

lint_test/prefer_guard_clause_test.dart
```diff
@@ -0,0 +1,288 @@
...
+  if (true) {
+    if (true) {
+      if (true) {
+        _doSomething();
+      }
+    }
+  }
+
+  _doSomething();
+}
```
yurii-prykhodko-solid, line 275: Don't think we should be linting this? If we convert it to the structure suggested by the lint (if x return), we'll miss the last `doSomething` call.

lib/src/lints/prefer_guard_clause/visitors/prefer_guard_clause_visitor.dart
```diff
@@ -0,0 +1,75 @@
...
+
+    switch (body.block.statements.length) {
+      case 1:
+        if (first is IfStatement) _handleIfStatement(first);
+        return;
+      case 2:
+        break;
+      case _:
+        if (last is ReturnStatement) {
+          last = body.block.statements[body.block.statements.length - 2];
```
yurii-prykhodko-solid, line 40: Why -2?

yurii-prykhodko-solid (changes requested):
The idea seems to mostly work, but I think we lack a clear definition of what we're trying to do here.

I'd look to these cases:
1. when `if` block is spanning the entire function body:
  ```dart
  void fn() {
    if (true) {
    ...
    }
  }
  // ->>
  void fn() {
    if (false) return;
    ...
  }
  ```

2. when `if-else` is spanning the entire function body

lib/src/lints/reverse_if_to_reduce_nesting/reverse_if_to_reduce_nesting_rule.dart
```diff
@@ -0,0 +1,67 @@
...
+/// #### BAD:
+///
+/// ```dart
+/// void func() {
+///   if(a){ //LINT
+///     if(b){ //LINT
+///       c;
+///     }
+///   }
+/// }
```
yurii-prykhodko-solid, line 21: Formatting:
```suggestion
///   if (a) { //LINT
///     if (b) { //LINT
///       c;
///     }
///   }
/// }
```

lib/src/lints/reverse_if_to_reduce_nesting/reverse_if_to_reduce_nesting_rule.dart
```diff
@@ -0,0 +1,67 @@
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:solid_lints/src/lints/reverse_if_to_reduce_nesting/visitors/reverse_if_to_reduce_nesting_visitor.dart';
+import 'package:solid_lints/src/models/rule_config.dart';
+import 'package:solid_lints/src/models/solid_lint_rule.dart';
+
+/// A `reverse_if_to_reduce_nesting` rule which promotes using guard clauses
+/// over nesting `if` statements
```
yurii-prykhodko-solid, line 8:
```suggestion
/// A rule which highlights `if` statements that span the entire body,
/// and suggests replacing them with a reversed boolean check with an early return.
```

lib/src/lints/reverse_if_to_reduce_nesting/reverse_if_to_reduce_nesting_rule.dart
```diff
@@ -0,0 +1,67 @@
...
+/// void func() {
+///   if(!a) return;
+///   if(!b) return;
+///   c;
+/// }
+/// ```
+class ReverseIfToReduceNestingRule extends SolidLintRule {
+  /// The [LintCode] of this lint rule that represents the error if
+  /// 'if' statements should be reversed
+  static const String lintName = 'reverse_if_to_reduce_nesting';
```
yurii-prykhodko-solid, line 36: A `guard` clause unfortunately has an [unconventional meaning in Dart,](https://dart.dev/language/branches#guard-clause) so we can't use that.
Also, not sure we need to put an accent on nesting here. Maybe something like this?
```suggestion
  static const String lintName = 'prefer_early_return';
```

lib/src/lints/reverse_if_to_reduce_nesting/visitors/if_statement_visitor.dart
```diff
@@ -0,0 +1,16 @@
...
+
+  /// All unnecessary if statements and conditional expressions.
+  Iterable<IfStatement> get nodes => _nodes;
+
+  @override
+  void visitIfStatement(IfStatement node) {
+    super.visitIfStatement(node);
+    _nodes.add(node);
+  }
+}
```
yurii-prykhodko-solid, line 16: Super nice!

lib/src/lints/reverse_if_to_reduce_nesting/visitors/reverse_if_to_reduce_nesting_visitor.dart
```diff
@@ -0,0 +1,54 @@
...
+        }
+      case _:
+        return;
+    }
+  }
+
+  void _handleIfStatement(IfStatement node) {
+    if (_isElseIfStatement(node)) return;
+    if (_hasElseStatement(node)) return;
+    if (!_containsIfStatement(node.thenStatement)) return;
```
yurii-prykhodko-solid, line 36: Could you explain why this is needed?
```suggestion
    final containsNestedIfStatement = _containsIfStatement(node.thenStatement);
    if (!containsNestedIfStatement) return;
```

yurii-prykhodko-solid (changes requested):
This looks like a great start for this lint!
We can expand on it in next iterations.
For now, just need to fixup a few minor issues, and clarify a few pieces.

yurii-prykhodko-solid (approved):
LGTM!

lint_test/prefer_early_return_test.dart
```diff
@@ -0,0 +1,237 @@
...
+    _doSomething();
+  }
+
+  return;
+}
+
+void nestedIf2() {
+  //expect_lint: prefer_early_return
+  if (true) {
+    if (true) {
```
vladimir-beloded, line 34: Can we show a warning here as well?

vladimir-beloded (changes requested):

lint_test/prefer_early_return_test.dart
4akloon: this block is illuminated by the warning in its entirety

illia-romanenko (approved):
LGTM - but missing changelog, I'll add it myself.

PR #160, opened 2024-04-17
Fix improper super calls in proper_super_calls.dart and add Disposable interface in proper_super_calls_test.dart

4akloon:


lint_test/proper_super_calls_test.dart
```diff
@@ -75,3 +75,7 @@ class MyClass {
   dispose() {}
   initState() {}
 }
+
+abstract interface class Disposable {
+  void dispose();
+}
```
solid-vovabeloded, line 81: Could you please add a comment here explaining what this testing is?

solid-vovabeloded (approved):

lint_test/proper_super_calls_test.dart
4akloon: Yep, added)

PR #161, opened 2024-04-17
149-new-lint-avoid_final_with_getter

4akloon:
added new avoid_final_with_getter rule


lint_test/avoid_final_with_getter_test.dart
```diff
@@ -0,0 +1,17 @@
...
+/// Check final private field with getter fail
+/// `avoid_final_with_getter`
+
+class Fail {
+  // expect_lint: avoid_final_with_getter
+  final int _myField = 0;
+
+  int get myField => _myField;
+
+  int get myField2 => _myField + 1;
```
solid-vovabeloded, line 12: I would say we need to create a separate class with such kind of getter to test we are not going to lint in this situation.

solid-vovabeloded (changes requested):
Great work so far, @4akloon! I've added a couple of suggestions here - please, take a look.

lint_test/avoid_final_with_getter_test.dart
4akloon: I will take it all to work

lint_test/avoid_final_with_getter_test.dart
```diff
@@ -0,0 +1,21 @@
+// ignore_for_file: type_annotate_public_apis, prefer_match_file_name, unused_local_variable
+
+/// Check final private field with getter fail
+/// `avoid_final_with_getter`
+
+class Fail {
+  // expect_lint: avoid_final_with_getter
+  final int _myField = 0;
+
+  int get myField => _myField;
```
solid-vovabeloded, line 10: If the getter name does not match the field name it does not report an issue. I think it would be great to handle this situation as well. What do you think?

solid-vovabeloded (changes requested):

lint_test/avoid_final_with_getter_test.dart
```diff
@@ -0,0 +1,27 @@
...
+class Skipped {
+  final int _myField = 0;
+
+  int get myField => _myField + 1; // it is not a getter for the field
+}
+
+class Skipped2 {
+  final int _myField = 0;
+
+  int get myFieldInt => _myField; // it is not a getter for the field too
```
solid-vovabeloded, line 22: I think in this case it should report the lint since it is a getter for the field.

solid-vovabeloded (changes requested):

lint_test/avoid_final_with_getter_test.dart
4akloon: yes, but I think it is necessary to additionally check whether there is another getter or variable with the same name.

solid-vovabeloded: I don't think it make sense to search for a getter with exactly the same name as a variable. If we have a private variable and a public getter(with any name) that returns this variable we should show lint. Also, I think it makes sense to highlight the getter, but not the variable, so the user can see why we are saying that this field should be just public.

lib/src/lints/avoid_final_with_getter/visitors/avoid_final_with_getter_visitor.dart
```diff
@@ -0,0 +1,34 @@
...
+  Iterable<MethodDeclaration> get getters => _getters;
+
+  @override
+  void visitMethodDeclaration(MethodDeclaration node) {
+    if (node
+        case MethodDeclaration(
+          isGetter: true,
+          declaredElement: ExecutableElement(
+            isAbstract: false,
+            isStatic: false,
```
solid-vovabeloded, line 21: Do we actually need to skip static getters? I think we may have the same situation when a static getter returns the static final private variable that either does not make any sense. I would remove this restriction.

solid-vovabeloded (changes requested):
Almost there, @4akloon! I've added one more suggestion and wanted to ask you to update the changelog with all the changes you've made so far.

lint_test/avoid_final_with_getter_test.dart
```diff
@@ -0,0 +1,28 @@
...
+  // expect_lint: avoid_final_with_getter
+  int get myField => _myField;
+}
+
+class Fail2 {
+  final int _myField = 0;
+
+  // expect_lint: avoid_final_with_getter
+  int get myFieldInt => _myField;
+}
```
solid-vovabeloded, line 18: Does it make sense to add a test for the static field and getter?

solid-vovabeloded (changes requested):
Please, update the changelog file as well as I mentioned before

lint_test/avoid_final_with_getter_test.dart
4akloon: yes, it does

solid-vovabeloded (approved):
LGTM! Great job, @4akloon.

PR #162, opened 2024-04-19
add exclude params support to avoid_returning_widgets rule

4akloon:


lib/src/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
```diff
@@ -92,11 +97,33 @@ class AvoidReturningWidgetsRule extends SolidLintRule {
...
+          :final String? className,
+        )) {
+      if (className == null) {
+        return true;
+      } else {
+        return returnType.hasIgnoredType(ignoredTypes: {className});
+      }
+    }
+
+    return false;
```
solid-vovabeloded, line 127:
```suggestion
    final excludedItem = config.parameters.exclude
        .firstWhereOrNull((e) => e.methodName == methodName);
   
   if (excludedItem == null) return false;
   
   final className = excludedItem.className;
   
   if (className == null) return false;
   
   return returnType.hasIgnoredType(ignoredTypes: {className});
```

I would refactor this part into something like this to make it a bit more readable. What do you think?

solid-vovabeloded (approved):
LGTM, thanks! Feel free to update changelog and merge it.

PR #171, opened 2024-05-02
I167

sufftea:


lib/src/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
```diff
@@ -96,7 +94,8 @@ class AvoidReturningWidgetsRule
 
       final isWidgetReturned = hasWidgetType(returnType);
 
-      final isIgnored = _shouldIgnore(node);
+      final isIgnored = config.parameters.matchMethod(node) ||
+          config.parameters.matchClass(node);
```
solid-vovabeloded, line 98: The class_name parameter means that the method_name in the class_name should not be processed by this lint. Not the whole class.

lib/src/lints/avoid_unused_parameters/visitors/avoid_unused_parameters_visitor.dart
```diff
@@ -48,6 +55,11 @@ class AvoidUnusedParametersVisitor extends RecursiveAstVisitor<void> {
         parameters.parameters.isEmpty) {
       return;
     }
+
+    if (ignoredEntities.matchClass(node)) {
```
solid-vovabeloded, line 59: I think the idea is to create a mechanism that will allow us to easilly reuse such code between different linter rules instead of just implementing it separately for each rule.

lib/src/lints/cyclomatic_complexity/models/cyclomatic_complexity_parameters.dart
```diff
@@ -1,19 +1,25 @@
-/// Cyclomatic complexity metric limits configuration.
+import 'package:solid_lints/src/lints/cyclomatic_complexity/models/max_cyclomatic_complexity_parameters.dart';
+import 'package:solid_lints/src/models/ignored_entities_model/ignored_entities_model.dart';
+
+/// Config parameters for the `cyclomatic_complexity` rule
 class CyclomaticComplexityParameters {
-  /// Threshold cyclomatic complexity level, exceeding it triggers a warning.
-  final int maxComplexity;
+  /// `max_complexity` configuration
+  final MaxCyclomaticComplexityParameters maxCyclomaticComplexity;
```
solid-vovabeloded, line 7: I do not think you need to create a separate class for extra params. Just go ahead and inherit this parameters class from the `IgnoredEntitiesModel`.

solid-vovabeloded (changes requested):

lib/src/lints/avoid_returning_widgets/avoid_returning_widgets_rule.dart
sufftea: It does work this way.

If you only specify the class name, then the entire class will be ignored, but if you specify both class and method name, then only this method in this class will be ignored.

```yaml
exclude:
  # excludes a matching method in a matching class
  - method_name: excludeMethod
    class_name: ExcludeClass
  # excludes a matching method anywhere
  - method_name: excludeFunction
  # excludes all methods within a matching class
  - class_name: ExcludeEntireClass
```

lib/src/lints/avoid_unused_parameters/visitors/avoid_unused_parameters_visitor.dart
sufftea: It has to be used slightly differently in different rules. I.e. in this one, the check has to be performed in the visitor, instead of directly in the `run` function, and `matchClass` and `matchMethod` have to be used separately. 

Most of the logic (parsing and matching against the list) is still inside the `IgnoredEntitiesModel`. Having two methods like that seemed like the minimal amount of code that is still flexible enough to adapt to different rules.

---

I just checked, and I think this particular rule can be implemented without the visitor. I can try that, but it won't change the way the check is performed here.

lib/src/lints/cyclomatic_complexity/models/cyclomatic_complexity_parameters.dart
sufftea: I think it's easier to implement `fromJson`  this way. Otherwise:
```dart
class CyclomaticComplexityParams extends IgnoredEntitiesModel {
  // ...  
  factory CyclomaticComplexityParams.fromJson(Map<String, Object?> json) {
    // With more properties, I'll have to go through all of them:
    final entities = IgnoredEntitiesModel.fromJson(json).entities;

    return CyclomaticComplexityParams(
      entities: entities,
      other: 1,
    );
  }
}
```

PR #186, opened 2024-12-04
Issue 167 1

StarovNikita:


lib/src/lints/avoid_unused_parameters/models/avoid_unused_parameters.dart
yurii-prykhodko-solid, line 1: move `models` -> `parameters`

lib/src/lints/avoid_returning_widgets/models/avoid_returning_widgets_parameters.dart
```diff
@@ -13,16 +13,11 @@ class AvoidReturningWidgetsParameters {
...
-      if (item is Map) {
-        exclude.add(AvoidReturningWidgetsExclude.fromJson(item));
-      }
-    }
     return AvoidReturningWidgetsParameters(
-      exclude: exclude,
+      exclude: ExcludedIdentifiersListParameter.fromJson(
+        excludeList:
+            json[ExcludedIdentifiersListParameter.excludeParameterName] as Iterable? ?? [],
+      ),
```
yurii-prykhodko-solid, line 20: 1. We should probably make the `excludeList` parameter positional instead of named;
2. This piece repeats a lot -- let's create a default one that would parse from `dynamic` maybe?
```suggestion
      exclude: ExcludedIdentifiersListParameter.defaultFromJson(json),
```

yurii-prykhodko-solid (changes requested):
Let's add at least one test per each rule, where we check whether the exclusion logic is working properly.

And please make sure the pipeline is passing.

lib/src/lints/avoid_unused_parameters/avoid_unused_parameters_rule.dart
```diff
@@ -91,12 +94,16 @@ class AvoidUnusedParametersRule extends SolidLintRule {
     ErrorReporter reporter,
     CustomLintContext context,
   ) {
-    context.registry.addCompilationUnit((node) {
-      final visitor = AvoidUnusedParametersVisitor();
-      node.accept(visitor);
+    context.registry.addDeclaration((node) {
+      final isIgnored = config.parameters.exclude.shouldIgnore(node);
+      
+      if (!isIgnored) {
```
yurii-prykhodko-solid, line 100: [Prefer early returns](https://lints.solid.software/docs/custom_lints/prefer_early_return)
```suggestion
      if (isIgnored) return;
```

lib/src/lints/cyclomatic_complexity/cyclomatic_complexity_rule.dart
```diff
@@ -52,13 +52,19 @@ class CyclomaticComplexityRule
     CustomLintContext context,
   ) {
     context.registry.addBlockFunctionBody((node) {
-      final visitor = CyclomaticComplexityFlowVisitor();
-      node.visitChildren(visitor);
+      context.registry.addDeclaration((declarationNode) {
+        final isIgnored =
+            config.parameters.exclude.shouldIgnore(declarationNode);
+        if (!isIgnored) {
```
yurii-prykhodko-solid, line 58: same as above

lib/src/lints/function_lines_of_code/models/function_lines_of_code_parameters.dart
```diff
@@ -1,26 +1,27 @@
...
+    required this.exclude,
   });
 
   /// Method for creating from json data
   factory FunctionLinesOfCodeParameters.fromJson(Map<String, Object?> json) =>
       FunctionLinesOfCodeParameters(
         maxLines: json['max_lines'] as int? ?? _defaultMaxLines,
-        excludeNames:
-            List<String>.from(json['excludeNames'] as Iterable? ?? []),
+        exclude: ExcludedIdentifiersListParameter.defaultFromJson(json),
```
yurii-prykhodko-solid, line 25: This is a breaking change. Changing it into the consistent naming probably does make sense, but we should bump the minor version and write it out in the CHANGELOG.

lib/src/lints/no_empty_block/models/no_empty_block_parameters.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:solid_lints/src/common/parameters/excluded_identifiers_list_parameter.dart';
+
+/// A data model class that represents the "avoid returning widgets" input
+/// parameters.
```
yurii-prykhodko-solid, line 4: Please make sure the comments are relevant.

lint_test/analysis_options.yaml
```diff
@@ -6,23 +6,43 @@ custom_lint:
...
       max_lines: 50
     - avoid_non_null_assertion
     - avoid_late_keyword:
       allow_initialized: true
     - avoid_global_state
-    - avoid_returning_widgets
+    - avoid_returning_widgets:
+      exclude:
+        - method_name: excludeWidgetMethod
+          class_name: ExcludeWidget
```
yurii-prykhodko-solid, line 28: Let's swap like this -- reads better IMO
```suggestion
        - class_name: ExcludeWidget
          method_name: excludeWidgetMethod
```

yurii-prykhodko-solid (changes requested):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.4
+- Added exclude classes, methods, functions into avoid_returning_widgets,avoid_unused_parameters, cyclomatic_complexity, function_lines_of_code, no_empty_bloc, number_of_parameters 
+- Changed exclude parameter name into unction lines of code from excludeNames to exclude
```
yurii-prykhodko-solid, line 3:
```suggestion
- BREAKING CHANGE: Renamed `excludeNames` parameter to `exclude` for `function_lines_of_code` lint.
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.4
+- Added exclude classes, methods, functions into avoid_returning_widgets,avoid_unused_parameters, cyclomatic_complexity, function_lines_of_code, no_empty_bloc, number_of_parameters
```
yurii-prykhodko-solid, line 2:
```suggestion
- Added `exclude` parameter for the following lints:
  - `avoid_returning_widgets`
  - ...
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.4
```
yurii-prykhodko-solid, line 1:
```suggestion
## 0.3.0
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+- Added exclude classes, methods, functions into avoid_returning_widgets,avoid_unused_parameters, cyclomatic_complexity, function_lines_of_code, no_empty_bloc, number_of_parameters 
+- Changed exclude parameter name into unction lines of code from excludeNames to exclude
+
 ## 0.2.4
 
 - Fixed an issue with `prefer_early_retrun` for throw expression
```
yurii-prykhodko-solid, line 7: Move this changelog into 0.3.0 -- these changes will go as a single release.

lib/src/lints/no_empty_block/models/no_empty_block_parameters.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:solid_lints/src/common/parameters/excluded_identifiers_list_parameter.dart';
+
+/// A data model class that represents the "no empty block" input
```
yurii-prykhodko-solid, line 3:
```suggestion
/// A data model class that represents the `no_empty_block` lint input
```

lint_test/function_lines_of_code_test/analysis_options.yaml
```diff
@@ -6,6 +6,6 @@ custom_lint:
   rules:
     - function_lines_of_code:
       max_lines: 5
-      excludeNames:
-        - "longFunctionExcluded"
-        - "longMethodExcluded"
+      exclude:
+         - method_name: longFunctionExcluded
```
yurii-prykhodko-solid, line 10: We want to allow specifying names as plain String lists. These will be unspecific, and work for both class names and method/function names.

It will also make the migration for existing users a bit easier. 
```suggestion
         - longFunctionExcluded
```

lib/src/common/parameters/excluded_identifiers_list_parameter.dart
```diff
@@ -33,12 +33,34 @@ class ExcludedIdentifiersListParameter {
...
+  }
+
   /// Returns whether the target node should be ignored during analysis.
   bool shouldIgnore(Declaration node) {
     final methodName = node.declaredElement?.name;
 
-    final excludedItem =
-        exclude.firstWhereOrNull((e) => e.methodName == methodName);
+    final excludedItem = exclude.firstWhereOrNull(
+      (e) => e.methodName == methodName || e.className == methodName,
```
yurii-prykhodko-solid, line 62: This removes some confusion.

And also highlights that we may not be comparing what we think we should be comparing.

We should check whether the declaration is a class before running `exclude.className == declarationName`; same for method names.

```suggestion
      (e) => e.methodName == declarationName || e.className == declarationName,
```

yurii-prykhodko-solid (changes requested):

lib/src/common/parameters/excluded_identifiers_list_parameter.dart
```diff
@@ -26,6 +26,39 @@ class ExcludedIdentifiersListParameter {
     for (final item in excludeList) {
       if (item is Map) {
         exclude.add(ExcludedIdentifierParameter.fromJson(item));
+      } else if (item is String) {
+        exclude.add(
+          ExcludedIdentifierParameter(
+            methodName: item,
+          ),
+        );
+      }
```
yurii-prykhodko-solid, line 35: For plain Strings -- I think it can be both classes _and_ methods.
Add a separate parameter, `declarationName` for such cases?

While we're at it, I think we should be careful about these exclusions being noo unspecific.

For example, I think the behavior should be like this:

```dart
exclude:
  - excluded

// lint
void excluded() {}

class A {
  // no lint
  void excluded() {}
}
```

lib/src/common/parameters/excluded_identifiers_list_parameter.dart
```diff
@@ -35,10 +68,11 @@ class ExcludedIdentifiersListParameter {
...
   /// Returns whether the target node should be ignored during analysis.
   bool shouldIgnore(Declaration node) {
-    final methodName = node.declaredElement?.name;
+    final declaredName = node.declaredElement?.name;
 
-    final excludedItem =
-        exclude.firstWhereOrNull((e) => e.methodName == methodName);
+    final excludedItem = exclude.firstWhereOrNull(
+      (e) => e.methodName == declaredName || e.className == declaredName,
+    );
```
yurii-prykhodko-solid, line 75: Check for declaration type before comparing. Right now we are checking classnames against method names and vice versa.

yurii-prykhodko-solid (changes requested):
Looking good, almost there. Please add tests for plain-string exclude.

lib/src/common/parameters/excluded_identifiers_list_parameter.dart
yurii-prykhodko-solid: Discussed separately -- it makes sense to keep. the current behavior.

lib/src/common/parameters/excluded_identifiers_list_parameter.dart
```diff
@@ -26,6 +26,39 @@ class ExcludedIdentifiersListParameter {
...
+
+    for (final item in excludeList) {
+      if (item is Map) {
+        exclude.add(ExcludedIdentifierParameter.fromJson(item));
+      } else if (item is String) {
+        exclude.add(
+          ExcludedIdentifierParameter(
+            declarationName: item,
+          ),
+        );
```
yurii-prykhodko-solid, line 61: You can replace all this with a call to another factory, to avoid duplication
```dart
  return ExcludedIdentifiersListParameter.fromJson(excludeList: excludeList);
```

yurii-prykhodko-solid (changes requested):

PR #188, opened 2024-12-05
issue-93. added new consider_making_a_member_privat rule

StarovNikita:


lint_test/consider_making_a_member_private_test.dart
```diff
@@ -0,0 +1,19 @@
...
+  void usedX() {}
+}
+
+class Y {
+  // no lint
+  void _y() {
+    final x = X();
+    x.usedX();
+  }
+}
```
yurii-prykhodko-solid, line 19: Add more tests for:
- fields (static, final, mutable)
- getters, setters
- methods (static/instance)
- constructors, factories

We should also consider this for non-class declarations (global functions, variables, constants) and class declarations as a whole.

yurii-prykhodko-solid (changes requested):

lint_test/consider_making_a_member_private_test.dart
```diff
@@ -0,0 +1,65 @@
...
+class Y {
+  // no lint
+  void _y() {
+    final x = X();
+    X.usedFactory();
+    x.usedMethodX();
+    usedGLobalFunction();
+    usedGlobalVariables;
+  }
+}
```
yurii-prykhodko-solid, line 65: This is not quite what the lint is supposed to do.

With the structure of the test, all these methods and functions could just as well be private too.

We should lint if the class/method/functions are not used _outside_ the file where they are declared.

PR #194, opened 2025-01-30
Fixed tests in test folders

Aygistov:
Tests in folders did not comply with the analysis_options.yaml in those same folders.


lib/src/lints/function_lines_of_code/function_lines_of_code_rule.dart
```diff
@@ -50,15 +50,25 @@ class FunctionLinesOfCodeRule
...
+      if (isIgnored) {
+        return;
+      }
+      checkNode(node);
+    }
 
-      if (isIgnored) return;
+    // Check for an anonymous function
+    void checkFunctionExpressionNode(FunctionExpression node) {
+      if (node.parent is FunctionDeclaration) {
```
yurii-prykhodko-solid, line 63: Add comment on why this check is here:
```suggestion
      // Prevent checking named functions twice
      if (node.parent is FunctionDeclaration) {
```

yurii-prykhodko-solid (approved):
LGTM!

Issue #124, opened 2024-02-02
Exclude `copyWith` method from `number_of_parameters` rule


ghost:

nathnaeld:
Additionally, provide an option to set separate limits on the `number_of_parameters` for named and unnamed methods.

Aygistov:
Resolves #196

PR #196, opened 2025-01-30
Exclude copyWith method from number_of_parameters rule

Aygistov:
Closes https://github.com/solid-software/solid_lints/issues/124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   - `number_of_parameters`
 - BREAKING CHANGE: Renamed `excludeNames` parameter to `exclude` for `function_lines_of_code` lint.
 - Fixed an issue with `prefer_early_retrun` for throw expression
-- Added `copyWith` to `exclude` section for `number_of_parameters` lint 
+- `number_of_parameters` lint: added `copyWith` to the default exclude list.
 
 ## 0.2.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
   - `no_empty_bloc`
   - `number_of_parameters`
 - BREAKING CHANGE: Renamed `excludeNames` parameter to `exclude` for `function_lines_of_code` lint.
-- Fixed an issue with `prefer_early_retrun` for throw expression 
+- Fixed an issue with `prefer_early_retrun` for throw expression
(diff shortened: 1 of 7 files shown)


yurii-prykhodko-solid (changes requested):
Please add tests.

lint_test/analysis_options.yaml
```diff
@@ -15,7 +15,8 @@ custom_lint:
       exclude:
         - class_name: Exclude
           method_name: avoidNumberOfParameters
-        - method_name: avoidNumberOfParameters          
+        - method_name: avoidNumberOfParameters
+        - method_name: copyWith
```
yurii-prykhodko-solid, line 19: We should check that the standard rule set works -- not a custom subset that we define here.

Let's create a separate nested package inside of lint_test, with its own `pubspec` and a clean `analysis_options` which would just import the standard rule set -- and then test on that.

yurii-prykhodko-solid (changes requested):
Also update the changelog, please.

lint_test/analysis_options.yaml
Aygistov: Made changes, please check.

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@
   - `no_empty_bloc`
   - `number_of_parameters`
 - BREAKING CHANGE: Renamed `excludeNames` parameter to `exclude` for `function_lines_of_code` lint.
-- Fixed an issue with `prefer_early_retrun` for throw expression 
+- Fixed an issue with `prefer_early_retrun` for throw expression
+- Added `copyWith` to `exclude` section for `number_of_parameters` lint
```
yurii-prykhodko-solid, line 11: - `number_of_parameters` lint: added `copyWith` to the default exclude list.

lint_test/analysis_options.yaml
```diff
@@ -15,7 +15,7 @@ custom_lint:
       exclude:
         - class_name: Exclude
           method_name: avoidNumberOfParameters
-        - method_name: avoidNumberOfParameters          
+        - method_name: avoidNumberOfParameters
```
yurii-prykhodko-solid, line 18: Remove the extra spaces
```suggestion
        - method_name: avoidNumberOfParameters
```

yurii-prykhodko-solid (approved):
Looking good! Just a couple nits.

Issue #190, opened 2024-12-19
Implement "use_nearest_context" rule


yurii-prykhodko-solid:
Right now we have an `unused_parameters` rule that forces us to rename all unused parameters to underscores.

In some cases, we not use context in widget builder functions initially, and add a usage later.
There is some potential of using an incorrect instance of `BuildContext` that way.

We want to add a lint that checks that we use `BuildContext` from _nearest_ available scope.

Example:

```dart
class SomeWidget extends StatefulWidget {
...
}

class _SomeWidgetState extends State<SomeWidget> {
  ...
  void _showDialog() {
    showModalBottomSheet(
      context: context,
      builder: (BuildContext _) {                <-- LINT -- this should be renamed to `context`
        final someProvider = context.watch<SomeProvider>();   <-- uses context from StatefulWidget instead of the builder, which causes errors.

        return const SizedBox.shrink();
      },
    );
  }
}
```

We also probably can provide a quick fix here.

PR #198, opened 2025-02-03
Issue 190. Implement use nearest context rule

Aygistov:


Aygistov:
Closes #190

lib/src/lints/use_nearest_context/models/use_nearest_context_parameters.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:solid_lints/src/common/parameters/excluded_identifiers_list_parameter.dart';
+
+/// A data model class that represents the "use nearest context" input
+/// parameters.
+class UseNearestContextParameters {
+  /// A list of methods that should be excluded from the lint.
+  final ExcludedIdentifiersListParameter exclude;
```
yurii-prykhodko-solid, line 7: Not quite sure how exclude should work in context of this method.
Besides I don't see it integrated into the rule implementation.

We should scrap this parameter for now.

Aygistov: Removed the exclude parameter.

yurii-prykhodko-solid: @Aygistov thank you!

Aygistov: Is it better to delete the entire UseNearestContextParameters class?

yurii-prykhodko-solid: Probably so -- I'll do that

yurii-prykhodko-solid (changes requested):
1. We don't seem to handle the cases where the nearest context is a class method:
```dart
class Test extends StatefulWidget {
  const Test({super.key});

  @override
  State<Test> createState() => _TestState();
}

class _TestState extends State<Test> {
  @override
  Widget build(BuildContext context) {
    showModalBottomSheet(
      context: context,
      builder: (BuildContext _) {
        return SizedBox.fromSize(size: context.size); // should be linted
      },
    );
    
    return const Placeholder();
  }
}
```

2. We should think about what to do with context from global keys. It's usually not a great idea to retrieve external context, so maybe worth to do this:
```dart

final key = GlobalKey<State>();

class Test extends StatelessWidget {
  const Test({super.key});

  @override
  Widget build(BuildContext context) {
    final outerContext = key.currentContext!;

    showModalBottomSheet(
      context: outerContext, // LINT
      builder: (BuildContext _) {
        return SizedBox.fromSize(size: outerContext.size); // LINT
      },
    );

    return const Placeholder();
  }
}

```

PR #207, opened 2025-04-30
Fix avoid using api

iimironchuk:
lint is working for extensions


yurii-prykhodko-solid (changes requested):
Please cherry-pick the `avoid_using_api` changes to a separate branch. This one has `prefer_match_file_name` changes bundled with it.

PR #212, opened 2025-06-05
Lint default constructor and named parameter (issue #126)

### LINKED ISSUE #126: avoid_using_api: Lint Default Constructor Only
While using the lint I noticed I missed an edge case, it is not possible to lint only the default constructor. it will always lint all usages of the class

Potential usage:

```yaml
custom_lint:
  rules:
    - avoid_using_api:
      entries:
        # Add parentheses to lint only default constructor, use `identifier` for named constructors
        - class_name: Border()
          source: package:flutter
          reason: Use `BorderDirectional` instead.
```

Not sure if this is the best way to implement it, and I could see it being confusing if named constructors (ie `class_name: Border.all()`) didn't also work. Thoughts?

I can implement this.

xJac0b:
Resolves #126 

Rule `avoid_using_api`:
- Lint only default constructor by using `()` identifier
- Lint usage of specific named parameter

### DIFF
--- a/lib/src/lints/avoid_using_api/avoid_using_api_linter.dart
+++ b/lib/src/lints/avoid_using_api/avoid_using_api_linter.dart
@@ -20,7 +20,7 @@ class AvoidUsingApiLinter {
   });
 
   /// The identifier for the default constructor
-  static const String defaultConstructorIdentifier = '()';
+  static const String _defaultConstructorIdentifier = '()';
 
   /// Access to the resolver for this lint context
   final CustomLintResolver resolver;
@@ -194,7 +194,7 @@ class AvoidUsingApiLinter {
     String className,
     String source,
   ) {
-    if (identifier == defaultConstructorIdentifier) {
+    if (identifier == _defaultConstructorIdentifier) {
       _banDefaultConstructor(className, source, entryCode);
       return;
     }
@@ -348,7 +348,7 @@ class AvoidUsingApiLinter {
     context.registry.addInstanceCreationExpression((node) {
       String? expectedConstructorName;
 
-      if (identifier != defaultConstructorIdentifier) {
+      if (identifier != _defaultConstructorIdentifier) {
         expectedConstructorName = identifier;
       }
 
--- a/lib/src/lints/avoid_using_api/avoid_using_api_linter.dart
+++ b/lib/src/lints/avoid_using_api/avoid_using_api_linter.dart
@@ -332,7 +332,12 @@ class AvoidUsingApiLinter {
         return;
       }
 
-      final sourcePath = enclosingElement.library2?.uri.toString() ?? '';
+      final libSource = enclosingElement.library2;
+      if (libSource == null) {
+        return;
+      }
+
+      final sourcePath = libSource.uri.toString();
       if (!_matchesSource(sourcePath, source)) {
         return;
       }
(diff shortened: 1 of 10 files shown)


lib/src/lints/avoid_using_api/avoid_using_api_linter.dart
```diff
@@ -280,4 +288,96 @@ class AvoidUsingApiLinter {
...
+          node.constructorName.type.element2?.library2?.uri.toString();
+      if (sourcePath == null || !_matchesSource(sourcePath, source)) {
+        return;
+      }
+
+      reporter.atNode(node, entryCode);
+    });
+  }
+
+  bool _hasNamedParameter(ArgumentList argumentList, String namedParameter) =>
```
solid-vovabeloded, line 378:
```suggestion
  bool _containsNamedParameter(ArgumentList argumentList, String namedParameter) =>
```
Would it be a little bit clearer?

lib/src/lints/avoid_using_api/avoid_using_api_linter.dart
```diff
@@ -280,4 +288,96 @@ class AvoidUsingApiLinter {
...
+      reporter.atNode(node.methodName, entryCode);
+    });
+
+    context.registry.addInstanceCreationExpression((node) {
+      final String? expectedConstructorName;
+      if (identifier == defaultConstructorIdentifier) {
+        expectedConstructorName = null;
+      } else {
+        expectedConstructorName = identifier;
+      }
```
solid-vovabeloded, line 353:
```suggestion
      String? expectedConstructorName;
      if (identifier != defaultConstructorIdentifier) {
        expectedConstructorName = identifier;
      }
```

lint_test/avoid_using_api/external_source/lib/external_source.dart
```diff
@@ -4,27 +4,35 @@
 class BannedCodeUsage {
   final String test4 = 'Hello World';
 
-  BannedCodeUsage();
+  BannedCodeUsage({this.parameter1, this.parameter2});
```
solid-vovabeloded, line 7: Since we are adding a new functionality that is actually not related to the `external_source` functionality, we should create a separate package like we do for other cases.

lint_test/avoid_using_api/named_parameter_ban/lib/named_parameter_ban_test.dart
```diff
@@ -0,0 +1,28 @@
...
+
+void testingBannedCodeLint() async {
+  // expect_lint: avoid_using_api
+  final bannedCodeUsage = BannedCodeUsage(parameter1: '');
+
+  // expect_lint: avoid_using_api
+  final test3 = BannedCodeUsage.test3(parameter1: '');
+
+  // expect_lint: avoid_using_api
+  final test5 = BannedCodeUsage.test5(parameter1: '');
```
solid-vovabeloded, line 13: What is the difference between these 2 test cases? That looks like it is testing the same piece of functionality here.

lint_test/avoid_using_api/named_parameter_ban/lib/named_parameter_ban_test.dart
```diff
@@ -0,0 +1,28 @@
...
+  );
+
+  final obj = BannedCodeUsage();
+  // expect_lint: avoid_using_api
+  obj.test(parameter1: '');
+
+  test(parameter1: '');
+
+  // expect_lint: avoid_using_api
+  2.banned(parameter1: '');
```
solid-vovabeloded, line 27: It would be great to add cases when we do not expect to have a lint for each test case, so we can be sure it does not generate false positives.

solid-vovabeloded (changes requested):

PR #214, opened 2025-06-06
Enable test/analysis_options.yaml (issue #192)

### LINKED ISSUE #192: analysis_options.yaml for test folder not working
Starting from 0.2.0 version. `analysis_options.yaml` file inside tests folder is ignored by linter.

xJac0b:
Resolves #192

### DIFF
--- a/lib/analysis_options.yaml
+++ b/lib/analysis_options.yaml
@@ -57,7 +57,7 @@ custom_lint:
     - double_literal_format
 
     - function_lines_of_code:
-      max_lines: 190
+      max_lines: 200
 
     - member_ordering:
       order:


lib/analysis_options.yaml
```diff
@@ -1,6 +1,4 @@
 analyzer:
-  plugins:
-    - custom_lint
```
solid-vovabeloded, line 3: I do not think we should change `analysis_options.yaml` here. The issue is about the `analysis_options_test.yaml` and its setup to work only for the `test` package of the end users.

solid-vovabeloded (changes requested):

PR #220, opened 2025-10-02
Add new rule: dont_create_a_return_var

### LINKED ISSUE #92: New rule: dont_create_a_return_var
BAD:
```dart
void x() {
  final y = 1;

  return y;
}
```

GOOD:
```dart
void x() {
  return 1;
}
```

tozpeak:
Resolves #92. Implemented new rule (dont_create_a_return_var) and tests for it.


lint_test/dont_create_a_return_var_test.dart
```diff
@@ -0,0 +1,125 @@
...
+/// right after declaration, which makes it bad.
+int returnVarTestReturnFollowsDeclaration() {
+  var a = 1;
+  final result = a;
+
+  //Some comment here
+
+  //expect_lint: dont_create_a_return_var
+  return result;
+}
```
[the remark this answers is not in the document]
tozpeak: @solid-dmytrotkachuk, regarding this test: it works as expected.

You can compare it with previous to see the difference I tried to emphasize here.
`returnVarTestCachedMutable` caches mutable variable before potential changing value. This is what I called "unpredictable", thus no lint there.
`returnVarTestReturnFollowsDeclaration` caches similar variable, but returns straight after caching, so it is unnecessary operation. I added comment line to check whether if linter understands that comments are not potential operations.

Seems like linter works just fine, when gemini sees comment as "something in-between" while semantically there's nothing.

lib/src/lints/dont_create_a_return_var/dont_create_a_return_var_rule.dart
```diff
@@ -0,0 +1,153 @@
...
+    if (expr case final BinaryExpression binExpr) {
+      return _isExpressionImmutable(binExpr.leftOperand) &&
+          _isExpressionImmutable(binExpr.rightOperand);
+    }
+
+    if (expr case final SimpleIdentifier identifier) {
+      return _isSimpleIdentifierImmutable(identifier);
+    }
+
+    return false;
```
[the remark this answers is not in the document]
tozpeak: I agree, it should look better with `switch`.

lib/analysis_options.yaml
```diff
@@ -56,6 +56,7 @@ custom_lint:
     - cyclomatic_complexity:
       max_complexity: 10
 
+    - dont_create_a_return_var
     - double_literal_format
```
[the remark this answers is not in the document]
tozpeak: Renamed in new version.

lint_test/analysis_options.yaml
```diff
@@ -89,3 +89,4 @@ custom_lint:
         - required
         - nullable
         - default
+    - avoid_unnecessary_return_variable
```
solid-dmytrotkachuk, line 92: I think we should move it to the other "avoid_" rules in this file now. That will make it easier to find in the find and will look cleaner.

lib/src/lints/dont_create_a_return_var/dont_create_a_return_var_rule.dart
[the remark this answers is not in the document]
tozpeak: Changed both methods to use switch statement.

lib/analysis_options.yaml
```diff
@@ -56,6 +56,7 @@ custom_lint:
     - cyclomatic_complexity:
       max_complexity: 10
 
+    - avoid_unnecessary_return_variable
     - double_literal_format
```
solid-dmytrotkachuk, line 60: I think we should group this with other avoid_unnecessary. It should improve the readability of the file and make it look cleaner.

image

lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    return false;
+  }
+
+  bool _isSimpleIdentifierImmutable(SimpleIdentifier identifier) {
+    switch (identifier.element) {
+      case final VariableElement2 variable
+          when variable.isFinal || variable.isConst:
+        return true;
```
solid-dmytrotkachuk, line 142: I think we can do the `return variable.ifFinal || variable.isConst` here instead of using when. It should make it easier to read.

lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
```diff
@@ -0,0 +1,156 @@
...
+
+      case final BinaryExpression binExpr:
+        return _isExpressionImmutable(binExpr.leftOperand) &&
+            _isExpressionImmutable(binExpr.rightOperand);
+
+      case final SimpleIdentifier identifier:
+        return _isSimpleIdentifierImmutable(identifier);
+    }
+
+    return false;
```
solid-dmytrotkachuk, line 135: We are already using switch here, so what do you think about adding `default:` case instead of separately returning false here?

solid-dmytrotkachuk (changes requested):
You are almost there, a couple more small changes, and it should be ready.

lint_test/analysis_options.yaml
tozpeak: This file, not the other one, is organized in what seems to be historical order. So there isn't a single group of "avoid_" rules, rather several clusters. It can be reorganized, but it seems out of scope.

lib/analysis_options.yaml
tozpeak: This one is fair though. Missed this while renaming.

lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
tozpeak: I deliberately changed it just to be consistent with neighboring method. The other one contains more complex evaluation, where having a `default:` clause was less readable than having single tail return.

I felt like this one can later get such logic as well, so they look better as a whole and I minimize following changes, just like having a trailing coma improves diff readability. This is a personal style preference only and I can change it to `default: return false;` if it fits style guide better.

lint_test/dont_create_a_return_var_test.dart
```diff
@@ -0,0 +1,125 @@
+// ignore_for_file: unused_local_variable, newline_before_return, no_empty_block
+
```
solid-dmytrotkachuk, line 2: This file name should also be changed to apply the new rule name

solid-dmytrotkachuk (approved):
LGTM, thanks!

PR #228, opened 2026-04-15
Migrate/avoid debug print in release mode

Islam-Shaaban-Ibrahim:
Migrate/avoid debug print in release mode


lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  void _check(AstNode node, SimpleIdentifier identifier) {
+    final element = identifier.element;
+    if (element == null) return;
+
+// Check the name
+    if (element.name == 'debugPrint') {
+      // Check if it's from the flutter/foundation path
+      final sourceUri = element.library?.uri.toString() ?? '';
```
solid-vovabeloded, line 44:
```suggestion

    if (element.name == 'debugPrint') {
      final sourceUri = element.library?.uri.toString() ?? '';
```

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,61 @@
...
+      final isFlutterFoundation =
+          sourceUri.contains('package:flutter/foundation.dart');
```
solid-vovabeloded:
```suggestion
      final isFlutterFoundation = sourceUri.contains(
        'package:flutter/foundation.dart',
      );
```
Also, it would be great to extract  'package:flutter/foundation.dart' into a constant.

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  bool _isWrappedInReleaseCheck(AstNode node) {
+    AstNode? parent = node.parent;
+    while (parent != null) {
+      if (parent is IfStatement) {
+        final expression = parent.expression;
+        final source = expression.toString();
+
+        if (source.contains('kReleaseMode') || source.contains('kDebugMode')) {
```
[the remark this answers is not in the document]
solid-vovabeloded: @Islam-Shaaban-Ibrahim, please, take a look.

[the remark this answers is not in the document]
solid-vovabeloded: Also, please, extract strings into constants.

test/avoid_debug_print_in_release_rule_test.dart
```diff
@@ -0,0 +1,88 @@
...
+      r'''
+import 'package:flutter/foundation.dart';
+void test() {
+  if (kDebugMode) {
+    debugPrint('Safe');
+  }
+}
+''',
+    );
+  }
```
solid-vovabeloded, line 134: I believe we need a few more test cases:

- assets in the following situations: 

```dart
import 'package:flutter/foundation.dart';
void test() {
  if (kReleaseMode) {
    debugPrint('Safe');
  }
}
``` 

and 


```dart
import 'package:flutter/foundation.dart';
void test() {
  if (!kDebugMode) {
    debugPrint('Safe');
  }
}
```
- does not assert if `debugPrint` is not from `package:flutter/foundation.dart`
- asserts if `debugPrint` is imported using `flutter/material.dart` or `flutter/cupertino.dart`

solid-vovabeloded (changes requested):
Well done. 
Please, take a look at the following suggestions and let me know what you think.

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,66 @@
...
+  bool _isWrappedInReleaseCheck(AstNode node) {
+    AstNode? parent = node.parent;
+    while (parent != null) {
+      if (parent is IfStatement) {
+        final expression = parent.expression;
+        final source = expression.toString();
+
+        if (source.contains(_kReleaseMode) || source.contains(_kDebugMode)) {
+          return true;
+        }
```
solid-vovabeloded: Are we going to handle `if (kReleaseMode)` and `if (kDebugMode)` the same way?

test/avoid_debug_print_in_release_rule_test.dart
```diff
@@ -0,0 +1,160 @@
...
+  ['a'].forEach(debugPrint);
+}
+''',
+      [lint(72, 10)],
+    );
+  }
+
+  /// Case: if (kReleaseMode) { ... } is considered safe/safe-guarded logic.
+  void test_does_not_report_kReleaseMode_guard() async {
+    await assertNoDiagnostics(
```
solid-vovabeloded: We should assert this case since this rule stands for guarding developers from printing in release mode.

test/avoid_debug_print_in_release_rule_test.dart
```diff
@@ -0,0 +1,160 @@
...
+      r'''
+import 'package:flutter/foundation.dart';
+
+void test() {
+  if (!kDebugMode) {
+    debugPrint('Safe');
+  }
+}
+''',
+    );
```
solid-vovabeloded: We should report an issue in this case.

solid-vovabeloded (changes requested):

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @override
+  void visitSimpleIdentifier(SimpleIdentifier node) {
+    /// Catch cases where debugPrint is passed as a reference:
+    /// final x = debugPrint;
+    if (node.parent is! MethodInvocation) {
+      _check(node, node);
+    }
+  }
```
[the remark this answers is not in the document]
solid-vovabeloded: @Islam-Shaaban-Ibrahim, have you tested this approach and case with `dabugPrint.call()`?

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,79 @@
...
+            condition.operator.lexeme == '!' &&
+            condition.operand is SimpleIdentifier) {
+          final operand = condition.operand as SimpleIdentifier;
+          if (operand.name == _kReleaseMode) {
+            return true;
+          }
+        }
+
+        // if (kDebugMode)
+        if (condition is SimpleIdentifier && condition.name == _kDebugMode) {
```
solid-vovabeloded: Please, remove unnecessary comments. According to clean code, if your code needs comments like this, that means there is room for readability improvement. In this concrete case, I'd better create 2 separate methods with readable names and extract these checks there instead of just adding comments.

solid-vovabeloded (changes requested):
Great job so far. I have the following thoughts here:

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,79 @@
...
+  bool _isWrappedInReleaseCheck(AstNode node) {
+    AstNode? parent = node.parent;
+
+    while (parent != null) {
+      if (parent is IfStatement) {
+        final condition = parent.expression;
+
+        // if (!kReleaseMode)
+        if (condition is PrefixExpression &&
+            condition.operator.lexeme == '!' &&
```
solid-vovabeloded: Would it be better to use `TokenType` here?

solid-vovabeloded (changes requested):

lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,89 @@
...
+    if (node.parent is! MethodInvocation) {
+      _check(node, node);
+    }
+  }
+
+  void _check(AstNode node, SimpleIdentifier identifier) {
+    final element = identifier.element;
+    if (element == null) return;
+
+// Check the name
```
solid-vovabeloded: This one is still here. I do not think it is useful.

solid-vovabeloded (approved):
Once small comment, but overall looks good to me, so I'm approving it right away. Feel free to merge once you resolve it.

PR #229, opened 2026-04-15
Migrate/proper super calls

Islam-Shaaban-Ibrahim:


lib/src/lints/avoid_debug_print_in_release/visitors/avoid_debug_print_in_release_visitor.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:solid_lints/src/lints/avoid_debug_print_in_release/avoid_debug_print_in_release_rule.dart';
+
+/// Visitor for [AvoidDebugPrintInReleaseRule].
+class AvoidDebugPrintInReleaseVisitor extends SimpleAstVisitor<void> {
```
solid-vovabeloded, line 7: I do not think it should be in this PR.

lib/src/lints/proper_super_calls/visitors/proper_super_calls_visitor.dart
```diff
@@ -0,0 +1,65 @@
...
+  void visitMethodDeclaration(MethodDeclaration node) {
+    final methodName = node.name.lexeme;
+    final body = node.body;
+
+    if ((methodName == _initState || methodName == _dispose) &&
+        body is BlockFunctionBody) {
+      final hasOverride = node.metadata.any(
+        (annotation) => annotation.name.name == _override,
+      );
+      if (!hasOverride) return;
```
solid-vovabeloded: It would be great if we could check if this method overrides Flutter's `State` class.

lib/src/lints/proper_super_calls/proper_super_calls_rule.dart
```diff
@@ -43,130 +43,50 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
-    final rule = RuleConfig(
-      name: lintName,
-      configs: configs,
-      problemMessage: (_) => 'Proper super calls issue',
-    );
-
-    return ProperSuperCallsRule._(rule);
-  }
+  @override
+  LintCode get diagnosticCode => _superInitStateCode;
```
solid-vovabeloded: I see we are repeating the same statements from rule to rule. It would be great if we could have something as we did before - `SolidLintRule`. It would allow us to get rid of boilerplate code and make our code cleaner.

lib/src/lints/proper_super_calls/proper_super_calls_rule.dart
```diff
@@ -43,130 +43,50 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
-        }
+    final visitor = ProperSuperCallsVisitor(
+      onViolation: (nameToken, {required bool isInitState}) {
+        // Access the reporter from the currentUnit
+        final reporter = context.currentUnit?.diagnosticReporter;
+
+        reporter?.atToken(
+          nameToken,
+          isInitState ? _superInitStateCode : _superDisposeCode,
+        );
```
solid-vovabeloded: Why are we passing this logic as a closure? Does it make sense to encapsulate it in the visitor as we do for all other rules?

test/proper_super_calls_rule_test.dart
```diff
@@ -0,0 +1,83 @@
...
+  void setUp() {
+    rule = ProperSuperCallsRule();
+    super.setUp();
+  }
+
+  @override
+  String get analysisRule => ProperSuperCallsRule.lintName;
+
+  /// Mocks the Flutter framework parts relevant to this lint.
+  String get _flutterBase => r'''
```
solid-vovabeloded:
```suggestion
  String get _flutterWidgetImplementation => r'''
```

test/proper_super_calls_rule_test.dart
```diff
@@ -0,0 +1,83 @@
...
+abstract class StatefulWidget {}
+abstract class State<T extends StatefulWidget> {
+  void initState() {}
+  void dispose() {}
+}
+''';
+
+  void test_initState_reports_when_not_first() async {
+    await assertDiagnostics(
+        _flutterBase +
```
solid-vovabeloded: Use dependency mocking instead, as you did for the avoid_debug_print rule tests.

test/proper_super_calls_rule_test.dart
```diff
@@ -0,0 +1,83 @@
...
+class MyWidgetState extends State<StatefulWidget> {
+  @override
+  void initState() {
+    print('Bad');
+    super.initState();
+  }
+}
+''',
+        [
+          lint(197, 9), // Highlights 'initState'
```
solid-vovabeloded:
```suggestion
          lint(197, 9), 
```
Please, remove unnecessary comments.

test/proper_super_calls_rule_test.dart
```diff
@@ -0,0 +1,83 @@
...
+
+  @override
+  void dispose() {
+    print('Good');
+    super.dispose();
+  }
+}
+''');
+  }
+}
```
solid-vovabeloded: I believe we miss some test cases here: 

- rule should not assert when we are not overriding the method.
- rule should assert when we do override, but omit explicit `@override` declaration.

Also, please think of edge cases and add more tests.

solid-vovabeloded (changes requested):

PR #230, opened 2026-04-15
migrated avoid non null assertion rule and tests

Islam-Shaaban-Ibrahim:


lib/src/lints/avoid_non_null_assertion/visitors/avoid_non_null_assertion_visitor.dart
```diff
@@ -0,0 +1,38 @@
...
+
+    final operand = node.operand;
+
+    if (operand is IndexExpression) {
+      final type = operand.target?.staticType;
+      final isInterface = type is InterfaceType;
+
+      final isMap = isInterface &&
+          (type.isDartCoreMap ||
+              type.allSupertypes.any((v) => v.isDartCoreMap));
```
solid-vovabeloded: Let's extract it into a separate method to improve readability.

lib/src/lints/avoid_non_null_assertion/avoid_non_null_assertion_rule.dart
```diff
@@ -37,56 +35,32 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
+class AvoidNonNullAssertionRule extends AnalysisRule {
+  /// Name of the lint
+  static const String lintName = 'avoid_non_null_assertion';
 
-  AvoidNonNullAssertionRule._(super.config);
+  /// Lint code used for suppression and reporting
+  static const LintCode code = LintCode(
+    lintName,
+    'Avoid using the bang operator. It may result in runtime exceptions.',
+  );
```
solid-vovabeloded: Can we make it private?

solid-vovabeloded (changes requested):
Overall, that looks pretty good. Just a few minor suggestions here.

lib/src/lints/avoid_non_null_assertion/avoid_non_null_assertion_rule.dart
solid-vovabeloded: @Islam-Shaaban-Ibrahim, can we apply it to all the rules we currently have migrated?

solid-vovabeloded (approved):

PR #234, opened 2026-04-23
Implementation of parameter parser

### LINKED ISSUE #245: Migrate parameter parsing

Dariaa14:
# Implementation:
Since the analysis_server_plugin currently lacks a direct way to retrieve the context root (noted by the TODO in the official source code), I explored two strategies for resolving analysis_options.yaml:

1. Plugin Register: I found this too restrictive, as it only functions correctly when 'dart analyze' is executed from the project root.

2. Rule Registration with Caching (Chosen): By resolving the configuration at the rule level, we can leverage the RuleContext to identify the file path currently under analysis. To prevent the performance overhead of re-parsing for every file, I implemented a caching layer. The YAML is only parsed once per 'dart analyze'; subsequent files retrieve the configuration from the cache.

# Research of other packages:
I researched the approach of others:

custom_lint and dart_code_metrics both utilize a custom server architecture built on the legacy analyzer_plugin package.
I found no existing public implementations for dynamic parameter injection for analysis_server_plugin users.

# Configuration Structure
Regarding our previous discussion on YAML structure: I tested the older formats used by other analyzers, but they were not recognized by the current server plugin. I have reverted to the structure specified in the official [Dart documentation](https://dart.dev/tools/analyzer-plugins), which I have verified as functional in this implementation.


lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+import 'package:analyzer/file_system/physical_file_system.dart';
+import 'package:solid_lints/src/common/parameter_parser/lint_options.dart';
+import 'package:yaml/yaml.dart';
+
+/// A global instance of [AnalysisOptionsLoader] for use across the plugin.
+final analysisLoader = AnalysisOptionsLoader();
+
+/// Loads and parses analysis options from a Dart project's YAML file.
+class AnalysisOptionsLoader {
+  Map<String, LintOptions> _rulesCache = {};
```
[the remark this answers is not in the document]
solid-vovabeloded: @Dariaa14, have you investigated the problem described here? Looks pretty relevant overall, but I'm not sure if changing analysis_options.yaml would restart the analysis server and reload the configuration.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+import 'dart:io' as io;
+import 'dart:io';
+import 'package:analyzer/analysis_rule/rule_context.dart';
+import 'package:analyzer/file_system/file_system.dart';
+import 'package:analyzer/file_system/physical_file_system.dart';
+import 'package:solid_lints/src/common/parameter_parser/lint_options.dart';
+import 'package:yaml/yaml.dart';
+
+/// A global instance of [AnalysisOptionsLoader] for use across the plugin.
+final analysisLoader = AnalysisOptionsLoader();
```
solid-vovabeloded: Please, avoid using top-level variables. I believe it would be better to instantiate this class during plugin creation and then inject it wherever we need.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  /// Retrieves the currently loaded lint rules.
+  Map<String, LintOptions> get rules => _rulesCache;
+
+  /// Loads lint rules from the analysis options file based
+  /// on the provided [RuleContext].
+  void loadRulesFromContext(RuleContext context) {
+    if (_rulesCache.isNotEmpty) {
+      return;
+    }
```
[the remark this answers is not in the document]
solid-vovabeloded: @Dariaa14, I think this one might be relevant as well.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+      }
+
+      final parent = dir.parent;
+
+      if (parent.path == dir.path) {
+        return null;
+      }
+
+      dir = parent;
+    }
```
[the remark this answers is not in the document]
solid-vovabeloded: @Dariaa14, please consider this one as well.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+      return;
+    }
+
+    final file = PhysicalResourceProvider.INSTANCE.getFile(yamlPath);
+
+    final rules = _getRules(file);
+    _rulesCache = rules;
+  }
+
+  String? _findNearestYamlUpwards(
```
solid-vovabeloded:
```suggestion
  String? _findNearestFileUpwards(
```
I believe it may be used for searching other files as well.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+    String filePath, {
+    String fileName = 'analysis_options.yaml',
+  }) {
+    final startFile = io.File(filePath);
+    io.Directory dir = startFile.parent;
+
+    while (true) {
```
solid-vovabeloded:
```suggestion
    while (dir.path != dir.parent.path) {
```
Does that make sense?

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+                map,
+                enabled: enabled,
+              );
+            }
+          }
+        }
+      }
+    }
+
+    return UnmodifiableMapView(rules);
```
solid-vovabeloded: If we are using `UnmodifiableMapView`, then we probably should have `_rulesCache` of `UnmodifiableMapView` type.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,118 @@
...
+
+/// A global instance of [AnalysisOptionsLoader] for use across the plugin.
+final analysisLoader = AnalysisOptionsLoader();
+
+/// Loads and parses analysis options from a Dart project's YAML file.
+class AnalysisOptionsLoader {
+  Map<String, LintOptions> _rulesCache = {};
+
+  /// Retrieves the currently loaded lint rules.
+  Map<String, LintOptions> get rules => _rulesCache;
```
solid-vovabeloded: Would it be great to have a method to get the options for the exact rule name?

lib/src/common/parameter_parser/lint_options.dart
```diff
@@ -0,0 +1,232 @@
...
+      : json = yaml;
+
+  /// Options with no [json]
+  const LintOptions.empty({required this.enabled}) : json = const {};
+
+  /// Whether the configuration enables/disables the lint rule.
+  final bool enabled;
+
+  /// Extra configurations for a [AnalysisRule].
+  final Map<String, Object?> json;
```
solid-vovabeloded: Please, declare fields and getters before the constructor.

solid-vovabeloded (changes requested):
Great start, @Dariaa14! Please, take a look at the following suggestions and let me know what you think.

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  /// Loads lint rules from the analysis options file based
+  /// on the provided [RuleContext].
+  void loadRulesFromContext(RuleContext context) {
+    if (context.allUnits.isEmpty) {
+      return;
+    }
+    final filePath = context.allUnits.first.file.path;
+    _loadRules(filePath);
+  }
```
solid-vovabeloded:
```suggestion
  LintOptions? getRuleOptions(RuleContext context, String ruleName) {
    final yamlPath = _findNearestFileUpwards(context.allUnits.first.file.path);
    
    if (yamlPath == null) return null;
    
    return _rulesCache[yamlPath]?[ruleName];
  }

  /// Loads lint rules from the analysis options file based
  /// on the provided [RuleContext].
  void loadRulesFromContext(RuleContext context) {
    if (context.allUnits.isEmpty) {
      return;
    }
    
    final filePath = context.allUnits.first.file.path;
    _loadRules(filePath);
  }
```

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,113 @@
...
+    final rules = _getRules(file);
+    _rulesCache[yamlPath] = rules;
+  }
+
+  String? _findNearestFileUpwards(
+    String filePath, {
+    String fileName = 'analysis_options.yaml',
+  }) {
+    final pathContext = PhysicalResourceProvider.INSTANCE.pathContext;
+    var dir = pathContext.dirname(filePath);
```
solid-vovabeloded: Avoid using `var`

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    final optionsString = analysisOptionsFile.readAsStringSync();
+    Object? yaml;
+    try {
+      yaml = loadYaml(optionsString) as Object?;
+    } catch (err) {
+      return {};
+    }
+    if (yaml is! Map) return {};
```
solid-vovabeloded:
```suggestion
    }
    
    if (yaml is! Map) return {};
```

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,113 @@
...
+              rules[ruleName] = LintOptions.empty(enabled: value);
+            } else if (value is Map) {
+              final map = Map<String, Object?>.from(value);
+
+              final enabled = map.remove('enabled') as bool? ?? true;
+
+              rules[ruleName] = LintOptions.fromYaml(
+                map,
+                enabled: enabled,
+              );
```
solid-vovabeloded: Why are we parsing `enabled` separately?

lib/src/lints/avoid_global_state/avoid_global_state_rule.dart
```diff
@@ -63,6 +66,10 @@ class AvoidGlobalStateRule extends AnalysisRule {
   ) {
     final visitor = AvoidGlobalStateVisitor(this);
 
+    _analysisLoader.loadRulesFromContext(context);
+    // To get the options of the rule:
+    // _analysisLoader.getRuleOptions(context, lintName);
```
solid-vovabeloded: I know this is just a usage example - leaving this comment for the future. We would need to create a separate class for our rules that would encapsulate this logic and just provide an API to get the rule settings.

solid-vovabeloded (changes requested):

lib/src/lints/avoid_global_state/avoid_global_state_rule.dart
```diff
@@ -63,6 +66,10 @@ class AvoidGlobalStateRule extends AnalysisRule {
   ) {
     final visitor = AvoidGlobalStateVisitor(this);
 
+    _analysisLoader.loadRulesOptionsFromContext(context);
+    // To get the options of the rule:
+    // _analysisLoader.getRuleOptions(context, lintName);
```
danylo-safonov-solid: not sure this is the correct place for this comment

lib/src/common/parameter_parser/analysis_options_loader.dart
```diff
@@ -0,0 +1,104 @@
...
+    if (cachedRules?.modificationStamp == modificationStamp) {
+      return;
+    }
+
+    final rules = _getRules(analysisOptionsFile);
+    _rulesCache[yamlPath] = CachedPackageRules(
+      modificationStamp: modificationStamp,
+      rules: rules,
+    );
+  }
```
danylo-safonov-solid:
```dart
  Map<String, Object?>? getRuleOptions(RuleContext context, String ruleName) =>
      _withNearestAnalysisOptionsFilePathForContext<Map<String, Object?>?>(
        context,
        (path) => _rulesCache[path]?.rules[ruleName],
      );

  /// Loads lint rules from the analysis options file for all rules
  /// using the provided [RuleContext].
  void loadRulesOptionsFromContext(RuleContext context) =>
      _withNearestAnalysisOptionsFilePathForContext(
        context,
        _loadRulesOptionsIfNewer,
      );

  T? _withNearestAnalysisOptionsFilePathForContext<T>(
    RuleContext context,
    T Function(String) f,
  ) {
    final packageRootPath = context.package?.root.path;
    if (packageRootPath == null) return null;

    final yamlPath = _findNearestAnalysisOptionsFilePath(packageRootPath);
    if (yamlPath == null) return null;

    return f(yamlPath);
  }

  void _loadRulesOptionsIfNewer(String yamlPath) {
    final analysisOptionsFile = _resourceProvider.getFile(yamlPath);
    final modificationStamp = analysisOptionsFile.modificationStamp;
    final cachedRules = _rulesCache[yamlPath];

    if (cachedRules?.modificationStamp == modificationStamp) {
      return;
    }

    final rules = _getRules(analysisOptionsFile);
    _rulesCache[yamlPath] = CachedPackageRules(
      modificationStamp: modificationStamp,
      rules: rules,
    );
  }
```

solid-danylosafonov (approved):

PR #237, opened 2026-06-03
Migrate avoid unnecessary return variable rule and its tests

daria-trusca-solid:
Migrated the `avoid_unnecessary_return_variable` rule and its corresponding tests to comply with the `analyzer` package.

### Changes
- **Rule Refactoring:** Updated the implementation of `avoid_unnecessary_return_variable` to use the current `analyzer` syntax.
- **Tests:** Updated and verified all associated test cases to ensure the rule triggers correctly and no regressions are introduced.


lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
```diff
@@ -32,109 +29,37 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
-Rewrite the variable evaluation into return statement instead.""",
-    );
+Rewrite the variable evaluation into return statement instead.""";
 
-    return AvoidUnnecessaryReturnVariableRule._(rule);
-  }
+  /// Lint code.
+  static const LintCode _code = LintCode(
+    lintName,
+    lintMessage,
```
danylo-safonov-solid: I think we can inline those the same way as I proposed in #238

lib/src/lints/avoid_unnecessary_return_variable/visitors/avoid_unnecessary_return_variable_visitor.dart
```diff
@@ -1,79 +1,79 @@
...
+      return;
+    }
 
-  /// After visiting holds info about whether there are any tokens
-  /// between variable declaration and return statement
-  bool get foundTokensBetweenDeclarationAndReturn =>
-      _foundTokensBetweenDeclarationAndReturn;
+    if (!element.isFinal && !element.isConst) {
+      return;
+    }
```
danylo-safonov-solid: nit - inconsistent braces. Since we are already omitting them in line 33, probably best to omit them throughout

```dart
    final expr = node.expression;
    if (expr is! SimpleIdentifier) return;

    final element = expr.element;
    if (element is! LocalVariableElement) return;

    if (!element.isFinal && !element.isConst) return;

    final block = node.parent;
    if (block == null) return;
```

lib/src/lints/avoid_unnecessary_return_variable/visitors/avoid_unnecessary_return_variable_visitor.dart
```diff
@@ -1,79 +1,79 @@
...
   }
 
-  void _checkTokensInBetween(
-    VariableDeclarationStatement variableDeclaration,
-    ReturnStatement returnStatement,
-  ) {
-    final tokenBeforeReturn =
-        _returnStatement.findPrevious(_returnStatement.beginToken);
+  bool _isSimpleIdentifierImmutable(SimpleIdentifier identifier) {
+    switch (identifier.element) {
```
danylo-safonov-solid: Not 100% sure this will work
```dart
    return switch (identifier.element) {
      ClassElement _ => true,
      final VariableElement variable ||
      GetterElement(:final VariableElement variable) =>
        variable.isFinal || variable.isConst,
      _ => false
    };
```

but this should

```dart
    return switch (identifier.element) {
      ClassElement _ => true,
      final VariableElement variable => variable.isFinal || variable.isConst,
      GetterElement(:final PropertyInducingElement variable) =>
        variable.isFinal || variable.isConst,
      _ => false
    };
```

lib/src/lints/avoid_unnecessary_return_variable/visitors/return_variable_usage_visitor.dart
```diff
@@ -0,0 +1,76 @@
...
+  VariableDeclaration? variableDeclaration;
+
+  /// After visiting holds info about whether there are any tokens
+  bool foundTokensBetweenDeclarationAndReturn = false;
+
+  /// The problem expects that exactly 1 mention of return variable.
+  /// VariableDeclarationStatement doesn't count when visiting SimpleIdentifier.
+  /// Any other amount of variable mentions implies that it is used somewhere
+  /// except return, so its existence is justified.
+  static const _badStatementCount = 1;
```
danylo-safonov-solid: nit - I'd rather have static as the top-most in the class

lib/src/lints/avoid_unnecessary_return_variable/visitors/return_variable_usage_visitor.dart
```diff
@@ -0,0 +1,76 @@
...
+
+  int _variableStatementCounter = 0;
+
+  /// Defines whether the variables is used in return statement only.
+  bool hasBadStatementCount() =>
```
danylo-safonov-solid:
```suggestion
  bool get hasBadStatementCount =>
```

danylo-safonov-solid (commented):
A few small suggestions, seems good otherwise

danylo-safonov-solid (approved):
lgtm

PR #238, opened 2026-06-03
Migrate avoid unnecessary set state rule and its tests

daria-trusca-solid:
Migrated the `avoid_unnecessary_set_state` rule and its corresponding tests to comply with the `analyzer` package.

### Changes
- **Rule Refactoring:** Updated the implementation of `avoid_unnecessary_set_state` to use the current analyzer syntax.
- **Tests:** Updated and verified all associated test cases to ensure the rule triggers correctly and no regressions are introduced.


lib/src/lints/avoid_unnecessary_setstate/visitors/avoid_unnecessary_set_state_visitor.dart
```diff
@@ -24,22 +24,26 @@
...
-  final _setStateInvocations = <MethodInvocation>[];
+  /// The rule associated with this visitor.
+  final AvoidUnnecessarySetStateRule rule;
+
+  /// Creates an instance of [AvoidUnnecessarySetStateVisitor].
+  AvoidUnnecessarySetStateVisitor(this.rule);
 
-  /// Unnecessary setState invocations
-  Iterable<MethodInvocation> get setStateInvocations => _setStateInvocations;
+  final _setStateInvocations = <MethodInvocation>[];
```
danylo-safonov-solid, line 43: should be before constructor
https://github.com/solid-software/solid_lints/blob/ad7ba630fc213fb3a4c962d49a0e16d89ef25261/lib/analysis_options.yaml#L65-L70

lib/src/lints/avoid_unnecessary_setstate/avoid_unnecessary_set_state_rule.dart
```diff
@@ -55,37 +55,36 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
 ///   if (mounted) setState(() => foo = 'bar');
 /// }
 /// ```
-class AvoidUnnecessarySetStateRule extends SolidLintRule {
-  /// The lint name of this lint rule that represents
-  /// the error whether we use setState in inappropriate way.
-  static const lintName = 'avoid_unnecessary_setstate';
+class AvoidUnnecessarySetStateRule extends AnalysisRule {
+  /// The name of the lint rule.
+  static const lintName = 'avoid_unnecessary_set_state';
```
danylo-safonov-solid: probably best to keep the current name https://lints.solid.software/docs/custom_lints/avoid_unnecessary_setstate

lib/src/lints/avoid_unnecessary_setstate/avoid_unnecessary_set_state_rule.dart
```diff
@@ -55,37 +55,36 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
-    final rule = RuleConfig(
-      name: lintName,
-      configs: configs,
-      problemMessage: (_) => 'Avoid calling unnecessary setState. '
-          'Consider changing the state directly.',
-    );
-    return AvoidUnnecessarySetStateRule._(rule);
-  }
+  /// The lint code for this rule.
+  static const LintCode _code = LintCode(
```
danylo-safonov-solid: nit
```suggestion
  static const _code = LintCode(
```

lib/src/lints/avoid_unnecessary_setstate/avoid_unnecessary_set_state_rule.dart
```diff
@@ -55,37 +55,36 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
-  /// The lint name of this lint rule that represents
-  /// the error whether we use setState in inappropriate way.
-  static const lintName = 'avoid_unnecessary_setstate';
+class AvoidUnnecessarySetStateRule extends AnalysisRule {
+  /// The name of the lint rule.
+  static const lintName = 'avoid_unnecessary_set_state';
 
-  AvoidUnnecessarySetStateRule._(super.config);
+  /// The message shown when the lint rule is triggered.
+  static const lintMessage = 'Avoid calling unnecessary setState. '
```
danylo-safonov-solid: Doesn't look like there are references to this variable outside of this class, it shouldn't be public. Also, both `lintName` and `lintMessage` are already exposed via `diagnosticCode` so that should be enough for testing

danylo-safonov-solid (changes requested):
A few final touches, otherwise looks good

danylo-safonov-solid (approved):
lgtm

danylo-safonov-solid (commented):
Is formatting broken in parent branch too? Should we make another PR to fix that (and any other CI checks if necessary)

daria-trusca-solid:
Yes, it looks like formatting was indeed broken on the parent branch. I just ran dart format . locally, and it modified quite a few files across the project. (I have the line length limit set to 80 characters)

PR #282, opened 2026-06-09
refactor: migrate avoid_unnecessary_type_assertions

### LINKED ISSUE #271: Migrate avoid_unnecessary_type_assertions

andrew-bekhiet-solid:
Fixes #271


test/lints/auto_test_lint_offsets.dart
```diff
@@ -0,0 +1,23 @@
...
+          lint(source.indexOf(codeFragment), codeFragment.length),
+      ];
+
+      await assertDiagnostics(source, expectedDiagnostics);
+    } finally {
+      _expectedCodeFragments = [];
+    }
+  }
+
+  String l(String code) {
```
solid-danylosafonov: Great stuff, but let's change the name to something more descriptive, maybe `expectLint`?

andrew-bekhiet-solid: I was thinking about keeping it as short as possible so the tested code has minimal impact on readability, what do you think?

solid-danylosafonov: I know it's a compromise, but someone new coming to work on the project might very well not know what `l` means here and will have to go through source code to understand that. But with `expectLint` naming it's obvious right away what it does.

danylo-safonov-solid:
Please avoid force pushing like that - it marks all the changes in the PR as unviewed, which I suspect is untrue in this case

danylo-safonov-solid (approved):

PR #283, opened 2026-06-15
fix: avoid_unnecessary_return_variable

### LINKED ISSUE #236: bug: false positive avoid_unnecessary_return_variable
Image

solid-illiaaihistov:
Closed https://github.com/solid-software/solid_lints/issues/236

### DIFF
--- a/lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
+++ b/lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
@@ -88,14 +88,14 @@ Rewrite the variable evaluation into return statement instead.""",
       return;
     }
 
-    final blockBody = statement.parent;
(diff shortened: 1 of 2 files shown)


lib/src/lints/avoid_unnecessary_return_variable/avoid_unnecessary_return_variable_rule.dart
```diff
@@ -88,14 +88,14 @@ Rewrite the variable evaluation into return statement instead.""",
       return;
     }
 
-    final blockBody = statement.parent;
-    if (blockBody == null) return;
+    final functionBody = statement.thisOrAncestorOfType<FunctionBody>();
```
danylo-safonov-solid, line 91: what about cases like:
```dart
final a = 3;
if (b) {
  return a;
}
```
`a` is still unnecessary here. Let's also add a test case like that

PR #292, opened 2026-06-17
fix: update avoid_unnecessary_return_variable

solid-illiaaihistov:
Closes https://github.com/solid-software/solid_lints/issues/236


solid-danylosafonov (approved):
Let's add another test, otherwise LGTM
```dart
  void
  test_does_not_report_if_return_is_cached_and_used_after_return_nested_block() async {
    await assertNoDiagnostics(r'''
int test(bool b) {
  final a = 3;
  if (b) {
    return a;
  }
  return a + 1;
}
''');
  }
```

PR #293, opened 2026-06-17
feat: add use_nearest_context rule

solid-illiaaihistov:
Closes https://github.com/solid-software/solid_lints/issues/190


lib/src/lints/use_nearest_context/visitors/use_nearest_context_visitor.dart
```diff
@@ -0,0 +1,74 @@
...
+    final element = node.element;
+    if (element is! LocalVariableElement) return false;
+
+    final nearestFunction = closestParam.parent?.parent;
+    if (nearestFunction is! FunctionExpression) return false;
+
+    final body = nearestFunction.body;
+    final declOffset = element.firstFragment.nameOffset;
+    if (declOffset == null) return false;
+    return declOffset >= body.offset && declOffset < body.end;
```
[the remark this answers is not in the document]
solid-illiaaihistov: The getter 'nameOffset' isn't defined for the type 'LocalVariableElement'.

lib/src/lints/use_nearest_context/fixes/replace_with_nearest_context_parameter_fix.dart
```diff
@@ -0,0 +1,68 @@
...
+          await builder.addDartFileEdit(file, (builder) {
+            builder.addSimpleReplacement(
+              parent.sourceRange,
+              parameterName,
+            );
+          });
+          return;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
solid-illiaaihistov: I intentionally separated the use of fixes for these two cases.

test/src/lints/use_nearest_context/use_nearest_context_rule_test.dart
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  Future<void> test_reports_on_outer_context_usage_with_nullable_nearest_context() async {
+    await assertDiagnostics(
+      r'''
+class BuildContext {
+  Object? get size => null;
+}
+class Widget {}
+void showModalBottomSheet({required BuildContext context, required Widget Function(BuildContext?) builder}) {}
```
solid-danylosafonov: we can extract that from all tests

```dart
@override
  void setUp() {
    final flutter = newPackage('flutter');
    flutter.addFile('lib/material.dart', r'''
class BuildContext {
  Object? get size => null;
}
class Widget {}
void showModalBottomSheet({required BuildContext context, required Widget Function(BuildContext?) builder}) {}
''');
```
And there are few other things that can be extracted there too scattered across other tests, like `Future.microtask`, etc

and then in each test we only need one line:
```dart
import 'package:flutter/material.dart';
```

test/src/lints/use_nearest_context/use_nearest_context_rule_test.dart
```diff
@@ -0,0 +1,423 @@
...
+  showModalBottomSheet(
+    context: context,
+    builder: (BuildContext _) {
+      final s = outerContext.size;
+      return Widget();
+    },
+  );
+}
+''',
+      [lint(345, 12)],
```
solid-danylosafonov: let's use this instead https://github.com/solid-software/solid_lints/blob/4e0f6ac7c4de7fa3cded7d1da2a7c7553587dd27/test/lints/auto_test_lint_offsets.dart#L50 

e.g.
https://github.com/solid-software/solid_lints/blob/4e0f6ac7c4de7fa3cded7d1da2a7c7553587dd27/test/lints/avoid_unnecessary_type_assertions/avoid_unnecessary_type_assertions_rule_test.dart#L167-L173

lib/src/lints/use_nearest_context/visitors/use_nearest_context_visitor.dart
```diff
@@ -0,0 +1,74 @@
...
+    if (!isBuildContext(node.staticType)) return;
+    if (_isPropertyOfOtherObject(node)) return;
+
+    final closestBuildContext = findClosestBuildContext(node);
+    if (closestBuildContext == null) return;
+    if (closestBuildContext.name?.lexeme != node.name) {
+      if (_isDeclaredInNearestScope(node, closestBuildContext)) return;
+
+      _rule.reportAtNode(node);
+    }
```
solid-danylosafonov:
```suggestion
    if (closestBuildContext.name?.lexeme == node.name) return;
    if (_isDeclaredInNearestScope(node, closestBuildContext)) return;

    _rule.reportAtNode(node);
```

lib/src/lints/use_nearest_context/visitors/use_nearest_context_visitor.dart
```diff
@@ -0,0 +1,73 @@
...
+    if (parent is PropertyAccess && node == parent.propertyName) {
+      var target = parent.target;
+      while (target is ParenthesizedExpression) {
+        target = target.expression;
+      }
+      return target is! ThisExpression && target is! SuperExpression;
+    }
+    return false;
+  }
```
solid-danylosafonov: Nit - maybe we should extract it to `utils/node_utils.dart`? - seems rather generic with potential for future reuse

lib/src/lints/use_nearest_context/visitors/use_nearest_context_visitor.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Returns `true` if [node] refers to a variable declared inside the body
+  /// of the function that owns [closestParam] (i.e. a local variable
+  /// in the same scope, like `final localCtx = innerContext;`).
+  bool _isDeclaredInNearestScope(
```
solid-danylosafonov: Nit:
1. We probably can extract it to utils as well
2. But, I'm not sure if the name reflects what this function does in the best way - maybe it should be something like `isDeclaredInSameFunction`? - that naming also leads me to think that it would look better as an extension for node so the call would look like `node.isDeclaredInSameFunction(as: param)`? WDYT?

solid-danylosafonov (approved):
LGTM, two optional suggestions

lib/src/lints/use_nearest_context/visitors/use_nearest_context_visitor.dart
solid-illiaaihistov: I think that's a great idea! I've updated it accordingly

PR #296, opened 2026-06-18
refactor: migrate function_lines_of_code rule

### LINKED ISSUE #251: Migrate function_lines_of_code

solid-illiaaihistov:
Closes https://github.com/solid-software/solid_lints/issues/251


test/src/lints/function_lines_of_code/function_lines_of_code_rule_test.dart
solid-danylosafonov, line 1: I think this would be a good case for table driven tests that I proposed in Slack last week

Something along the lines of:
```dart
  String testFunctionText(TestCase case)  {
    final comment = case.comments ? '\n// This is a single-line comment.\n' : '';
    final multilineComment = case.comments ? '\n/*\n  * This is a multi-line comment.\n  */\n' : '';

    return '''
int function() {${comment}
  var i = 0;
  ${repeat('i++';, case.lines - 2)\n}${multilineComment}
  
  return i;
}
''');
  }

String repeat(String string, int times) => List.generate(times, (_) => string).join();

typedef TestCase = ({int lines, bool comments});
enum ExpectedResult {Pass, Fail};

Map<TestCase, ExpectedResult> testTable = {
  (lines: 4, comments: true): Pass,
  (lines: 3, comments: false): Pass,
  ...
};

for (final (case, result) in testTable) {
  final text = testFunctionText(case);

  switch (result) {
    case Pass:
      assertNoDiagnostics(text);
    case Fail:
      assertAutoDiagnostics('${expectLint(text)}');
  }
}
```

Some of those probably best extracted to some utils, and maybe a basic framework for table tests too, since we are likely to reuse it

lib/src/lints/function_lines_of_code/visitors/function_lines_of_code_rule_visitor.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  @override
+  void visitFunctionExpression(FunctionExpression node) {
+    if (node.parent is! FunctionDeclaration) {
+      _checkNode(node);
+    }
+    super.visitFunctionExpression(node);
+  }
+
+  void _checkNode(AstNode node) {
```
solid-danylosafonov, line 43:
```dart
  void _checkNode(AstNode node) {
    final currentUnit = _context.currentUnit;
    if (currentUnit == null) return;

    final lineInfo = currentUnit.unit.lineInfo;
    final visitor = FunctionLinesOfCodeVisitor(lineInfo);
    node.visitChildren(visitor);
    if (visitor.linesWithCode.length <= _parameters.maxLines) return;

    final reporter = currentUnit.diagnosticReporter;
    if (node is! AnnotatedNode) {
      reporter.atNode(
        node,
        _rule.diagnosticCode,
        arguments: [_parameters.maxLines],
      );

      return;
    }

    final startOffset = node.firstTokenAfterCommentAndMetadata.offset;
    final lengthDifference = startOffset - node.offset;

    reporter.atOffset(
      offset: startOffset,
      length: node.length - lengthDifference,
      diagnosticCode: _rule.diagnosticCode,
      arguments: [_parameters.maxLines],
    );
  }
```

test/src/lints/function_lines_of_code/function_lines_of_code_rule_test.dart
solid-illiaaihistov: Great idea! Let me know what you think of this implementation.

test/src/utils/code_generators.dart
```diff
@@ -0,0 +1,5 @@
+/// Repeats [string] [times] times, joining with newlines.
+///
+/// Useful for generating test Dart code with a specific number of lines.
+String repeatLines(String string, int times) =>
```
solid-danylosafonov: Nit:
On second thought - let's make it an extension, I think `'i++'.repeat(2)` looks cleaner

```suggestion
String repeatLines(int times) =>
```

test/src/utils/table_driven_rule_test.dart
```diff
@@ -0,0 +1,53 @@
...
+
+      try {
+        switch (expected) {
+          case ExpectedResult.pass:
+            await assertNoDiagnostics(source);
+          case ExpectedResult.fail:
+            final marked = source.replaceFirst(
+              lintTarget,
+              expectLint(lintTarget),
+            );
```
solid-danylosafonov: I'm a bit confused - why don't we make `generateCode` return just a String? it can call `expectLint` internally

solid-danylosafonov (approved):
lgtm, two small things commented

test/src/utils/table_driven_rule_test.dart
solid-illiaaihistov: I think this makes sense.

solid-illiaaihistov: I think it would be good if the generator returned the resulting string directly.

PR #299, opened 2026-06-24
refactor: migrate no_empty_block

### LINKED ISSUE #254: Migrate no_empty_block

solid-illiaaihistov:
Closes #254


test/src/lints/no_empty_block/no_empty_block_rule_test.dart
```diff
@@ -0,0 +1,198 @@
...
+      testPackageRootPath,
+      '''${analysisOptionsContent(rules: [rule.name])}
+plugins:
+  solid_lints:
+    diagnostics:
+      no_empty_block:
+        exclude:
+          - method_name: excludeMethod
+          - class_name: Exclude
+            method_name: excludeMethod
```
solid-danylosafonov: I think we should rename it so that we can be sure that it's not ignored because of the above `- method_name: excludeMethod`
```suggestion
            method_name: excludeClassMethod
```

solid-danylosafonov (approved):
lgtm, one minor suggestion

PR #322, opened 2026-07-20
Fix GitHub Actions workflow

### LINKED ISSUE #307: Fix GitHub Actions workflow
To fix the auto-formatting issue in GitHub Actions before the release, I suggest running `dart format .` on the entire project.

solid-illiaaihistov:
Closes #307


lib/src/lints/no_empty_block/no_empty_block_rule.dart
```diff
@@ -1,3 +1,4 @@
+// ignore_for_file: todo
```
danylo-safonov-solid: looks like we can do inline `// ignore: todo` and it doesn't affect the doc comment

danylo-safonov-solid (approved):
One small thing, otherwise LGTM

PR #337, opened 2026-07-28
docs: improve avoid_duplicate_code rule and parameter documentation

solid-illiaaihistov:
## Summary by CodeRabbit

* **Documentation**
  * Expanded guidance for duplicate-code detection, including clone classification, token-based analysis, nested clone suppression, best-effort cross-file scanning, on-disk caching, and false-positive avoidance strategies.
  * Clarified duplicate-code configuration options (token thresholds, identifier/literal ignoring, and block checking) with improved explanations and examples.
  * Refreshed the `avoid_duplicate_code` example configuration: updated `minTokens`, enabled `checkBlocks`, and revised `exclude` entries.

### DIFF
--- a/lib/src/lints/avoid_duplicate_code/avoid_duplicate_code_rule.dart
+++ b/lib/src/lints/avoid_duplicate_code/avoid_duplicate_code_rule.dart
@@ -73,10 +73,12 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
 /// * **Coincidental Duplication:** Not all duplication is harmful. If two
 ///   identical code blocks serve completely distinct business purposes, merging
 ///   them creates a forced dependency. This can lead to
-///   [premature abstraction](https://medium.com/@ricrivero3/premature-abstraction-is-the-root-of-all-evil-7309762c0635)
-///   and [tight coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)).
+///   [hasty abstractions][aha] and [tight coupling][coupling].
 ///
 /// :::
+///
+/// [aha]: https://en.wikipedia.org/wiki/Don't_repeat_yourself#AHA
+/// [coupling]: https://en.wikipedia.org/wiki/Coupling_(computer_programming)
 ///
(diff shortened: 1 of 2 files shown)


lib/src/lints/avoid_duplicate_code/avoid_duplicate_code_rule.dart
```diff
@@ -54,29 +53,39 @@ import 'package:solid_lints/src/models/solid_lint_rule.dart';
...
+/// Because this rule analyzes the AST structure, you may encounter warnings in
+/// scenarios where extracting the code is actually undesirable:
+///
+/// * **Boilerplate Code:** Repetitive structures like form initializers, DTO
+///   mappers, or standard lifecycle hooks (e.g., `initState`, `dispose`) often
+///   share identical AST flows by necessity.
+/// * **Coincidental Duplication:** Not all duplication is harmful. If two
+///   identical code blocks serve completely distinct business purposes, merging
+///   them creates a forced dependency. This can lead to
+///   [premature abstraction](https://medium.com/@ricrivero3/premature-abstraction-is-the-root-of-all-evil-7309762c0635)
```
illia-romanenko: I don't like this reference, it's not as informative and feels a bit biased? We should link to research or wikipedia.

solid-illiaaihistov: Updated:
```dart
/// * **Coincidental Duplication:** Not all duplication is harmful. If two
///   identical code blocks serve completely distinct business purposes, merging
///   them creates a forced dependency. This can lead to
///   [hasty abstractions][aha] and [tight coupling][coupling].
///
/// :::
///
/// [aha]: https://en.wikipedia.org/wiki/Don't_repeat_yourself#AHA
/// [coupling]: https://en.wikipedia.org/wiki/Coupling_(computer_programming)
```
[https://en.wikipedia.org/wiki/Don't_repeat_yourself#AHA](https://en.wikipedia.org/wiki/Don't_repeat_yourself#AHA)

